PR #9, opened 2016-09-12
added nofish_Realtime clock.lua

nofishonfriday:
Hi ReaTeam,

I've added my clock script. 
Since it's my first contribution please check if I did the header correctly.

Thanks,
nofish

### DIFF
--- a/Various/nofish_Realtime clock.lua
+++ b/Various/nofish_Realtime clock.lua
@@ -1,5 +1,17 @@
--- Realtime clock
--- see http://forum.cockos.com/showthread.php?t=165884
+--[[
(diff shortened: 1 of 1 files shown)


nofishonfriday:
That was fast, thanks.
And thanks also for doing ReaPack btw., it's brilliant.

cfillion:
There was a missing `*` at line 7 for the About tag, I've added it in 910260e7dfe71834a28a7dc5fe7ece4395adf430.
(You can use whatever comment style you prefer, but reapack-index's parser need it to be the same across all lines of a single tag)

Thank you for uploading your script!

nofishonfriday:
Got it, noted for the next time. :)

PR #10, opened 2016-09-15
Create js_Zoom MIDI editor to 5 measures at mouse position.lua

juliansader:


cfillion:
Mhh merging was blocked, sorry for that... The repo settings should be back to normal now.

PR #11, opened 2016-10-15
Item Comps

JamesHEure:
Let me know if I need additional meta data. The Comps display needs to @import the Comp gfx and string functions.

Wasn't sure if I should make a new folder or not... (wasn't sure where it fits) If you'd rather this not create a new folder let me know.


cfillion:
Please add a `JamesH_` or similar author prefix to the filenames, to follow the naming convention documented in the README file of this repo.

I recommend using a `@description` close to the file name (without author prefix/file extension) so it's [easier](https://i.imgur.com/OvHc7jZ.png) to figure out how to find it in the Action List. `@about` is better suited for more verbose descriptions.

I suggest hiding `Comp gfx and string functions.eel`  from the Action List: you can do that by adding "`@provides [nomain] .`" in its header. I suppose other scripts in the future will reuse that file?

Pinging @X-Raym and @MichaelPilyavskiy for the directory name. They did the script categorization. The dash should probably be a space to match with the other existing categories though.

JamesHEure:
Thanks for the help! I'm really good as missing the obvious sometimes- like the filename prefix convention.

I found a bug that I'm trying to sort through and would prefer to fix it before this pull request is completed. I'll fix that, (or hack it into irrelevance)  and I will make the changes suggested. 

I guess this could go in the Item Editing Folder instead? it's more of a "takes" thing - not sure what will be the most clear for the end user.

It's a big unwieldy script - trying to clean it up :)

X-Raym:
@cfillion Hmm Item Comps is in between items properties and items editing, as it is not really an item properties, and it involves different items but linked to one position, maybe we can have an a didcated folder as proposed in the feature request. Under the folder naming convention, it should be named **Items Comp**.

@JamesHEure  Thanks for your participation here, I hope you like contribute to this repo :)

X-Raym:
@JamesHEure does your comp system use the native Comp Takes system ?
If yes, then the folder may be better named Comp Takes (just like the actions in action list).

JamesHEure:
@X-Raym  No, it doesn't use the native Comp Takes system. But I think "Comp Takes" is actually a better name, seems a bit more clear.

I fixed the bug I had found (related to take offsets) - so I'm almost ready to go!  I need to code in  some logic regarding the new take lane behavior with selected / grouped items which was introduced in 5.24.

X-Raym:
@JamesHEure Ok, but if your system doesn't use the native comp takes system, maybe we need to find another folder name to avoid mis-interpretation. People may want actions related to native comp system.
I propose to name it just Comp then, or Comping 👍

JamesHEure:
hmmm... I think just putting it under "Various" makes the most sense., since it relates to and effects Various things- or I'm not sure. Change it to whatever. You can simply comp takes with it or you can even A/B crossfades between two items!

so it's takes, items, and item fades. (and I suppose take offsets as well) 

I can't change the folder in this pull request (or at least didn't figure out how _ i guess I can start over? ) oh boy!

I fixed my bugs (created new ones and fixed those as well ha ha) and also managed to make the script not dependent on having SWS installed. (this was not an easy task - but something I've wanted to accomplish since the start - over a year ago)

Also had to reinstall Windows over the weekend cause I have a failing hard  drive,. So much fun! :(

I committed my changes, I do need to make better documentation - my idea is to put a 'help' button in the gui, I think it's ready though!

JamesHEure:
@X-Raym @cfillion 

I can't change the directory - when I edit it doesn't seem to be an option.  Should I close this and reopen another pull request?

X-Raym:
@JamesHEure I was thinking...
It seems ti me that you have a lot of scripts.
Rather than having to create pull request at every modifications, cause it will happen a lot ( 😺 )
I can propose you to either.
- Have your own repo
- Be invited in this repo as a member

The first solution give you more freedom and independence on your files management, but need to use reapack to index your repo after a commit (nothing complicated once it is setup).
The second solution is cool too.

What do you think ?

JamesHEure:
@X-Raym  I do plan to manage my own repo. Will looking into setting up reapack.

With this particular script, I decided to get it up on ReaTeam (at least temporarily) to potentially reach more users - it's something that I feel needs a lot of user feedback in order to make it better.

X-Raym:
@JamesHEure Well you can have your own repo AND be part of ReaTeam.
I will send you an invitation 👍

JamesHEure:
OK! figured out how to change the directory! yay!

X-Raym:
@JamesHEure you should have recieved an [invitation](https://github.com/ReaTeam/ReaScripts/invitations).
Let us know when this commit will be ready so that we can merge. Then you will be able to make more modifications right on the main repo.

PR #13, opened 2016-12-14
New 'Randomize pan of selected items' script

Nagasaki45:
It's my first time using reascripts / writing lua / contributing here. Any comment to improve this PR is more than welcome!

### DIFF
--- a/Items Editing/Nagasaki45_Randomize pan of selected items.lua
+++ b/Items Editing/Nagasaki45_Randomize pan of selected items.lua
@@ -1,0 +1,11 @@
+-- @description Randomize pan of selected items
+-- @version 0.1
+
+selected_item_count = reaper.CountSelectedMediaItems(0)
+for i = 0, selected_item_count - 1 do
+    item = reaper.GetSelectedMediaItem(0, i)
+    item_take = reaper.GetActiveTake(item)
+    random_pan = math.random() * 2 - 1
+    reaper.SetMediaItemTakeInfo_Value(item_take, 'D_PAN', random_pan)
+    reaper.UpdateItemInProject(item)
+end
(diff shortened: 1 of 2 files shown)


cfillion:
Looks good! I suggest adding the [`@author`](https://github.com/cfillion/reapack-index/wiki/Packaging-Documentation#author) tag in the header with your username (or name).

cfillion:
Thanks the contribution! I suggest adding the [`@author`](https://github.com/cfillion/reapack-index/wiki/Packaging-Documentation#author) tag in the header with your username (or name).

Since pan is tied to a specific take and not to the whole item, maybe this script would be best moved in the `Takes Properties` category and renamed as such? There is a similar script there for randomizing take's pitch from spk77 named "Randomize active takes pitch in selected items".

**EDIT:** There is also another "Pan active takes randomly" script there, but that one overwrites the take's pan envelope, and so it's quite different than yours.

Nagasaki45:
Thanks @cfillion!
I wasn't really aware of the difference between items and takes. I will move it to the correct folder.
I didn't add the author tag because I'm not active on reaper formus. Adding my github username will be OK?

cfillion:
Yes, GitHub username is fine. Users will be able to see that from various places where the full filename is not shown, rather than seeing "Unknown" or nothing.

Nagasaki45:
Hi @cfillion. Author added and the script was moved to the correct dir and rename to follow a similar name to the script of spk77.

cfillion:
Thank you! Could you edit `@description` to match the new filename as well?

```
@description Randomize active takes pan in selected items
```

This way it will have a similar name in both ReaPack (which uses description starting with v1.1) and REAPER's Action List, making it easier to find.

Nagasaki45:
Oh, sure, thanks for putting the effort to explain simple things to a newby :-)

PR #14, opened 2016-12-23
Added one script

Landeplage:
It takes the selected takes names and put the first character at the end.

I made it for a specific case where I had 130 items named "3F#", "5C", "6D#", and so on. I wanted to move the number to the end of the name.


X-Raym:
@Landeplage 
This is a very particular script ^^

PR #16, opened 2016-12-23
Mordi_Set selected items active takes name to track name.lua

Landeplage:
Sorry about all these pulls. I won't make any more scripts tonight. :p

### DIFF
--- a/Mordi_Set selected items active takes name to track name.lua
+++ b/Mordi_Set selected items active takes name to track name.lua
@@ -1,0 +1,37 @@
+-- @version 1.0
+-- @author Mordi
+-- @changelog
+--  This script will take the selected items' active take
+--  and rename it to whatever the track it is on is called.
+--
+--  Made by Mordi, Dec 2016
+
+-- Begin undo-block
+reaper.Undo_BeginBlock2(0)
+
+-- Print function
+function print(str)
+  reaper.ShowConsoleMsg(tostring(str) .. "\n")
+end
+
+-- Loop through all selected items
+for i = 0, reaper.CountSelectedMediaItems()-1 do
+  
+  -- Get item
+  item = reaper.GetSelectedMediaItem(0, i)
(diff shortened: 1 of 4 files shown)


cfillion:
Thanks for uploading those scripts! No need to be sorry for making a few pull requests: it's the entire purpose of this repository. I have two suggestions regarding the header:

1. I recommend using the [`@about`](https://github.com/cfillion/reapack-index/wiki/Packaging-Documentation#about) tag rather than `@changelog` for describing the file's general purpose. It's better suited for that with much more formatting capability using Markdown. (Furthermore ReaPack v1.0 has a bug causing it to crash when displaying a changelog if it contains empty lines like in yours.)
2. You could add a [`@description`](https://github.com/cfillion/reapack-index/wiki/Packaging-Documentation#description) with the name of the file without author prefix and extension to make it prettier in the list (as seen in the screenshot below).

```
-- @description Set selected items active takes name to track name
-- @version 1.0
-- @author Mordi
-- @about
--   This script will take the selected items' active take
--   and rename it to whatever the track it is on is called.
--
-- Made by Mordi, Dec 2016
```
With both changes it would look like this:

Screenshot

Landeplage:
Aha. My headers are based on the norm from months ago. I'll see if I can change them and reupload somehow.

Landeplage:
I added the new headers. Let me know if they are wrong, and I'll change them again. :p

cfillion:
Thanks! I merged this PR manually due to a conflict between your master branch and ours. The easiest way to avoid that in the future is to use the Create/Edit file buttons directly on the ReaTeam repo instead of those on the personal fork (the original contributing instructions I wrote were wrong in that regard).

Also there was an extra copy of the script at the root directory, I removed it to keep only the one in Takes Properties.

The new script and changes will be published in a few minutes if they aren't already! 🎅

PR #19, opened 2017-01-09
Duplicate notes

jalovatt:


X-Raym:
@jalovatt Hi ! Why all these scripts get deleted ? Do you have a custom repo now ?

jalovatt:
No, the original way I had written/uploaded them was really messy, didn't share any code, etc.
@cfillion helped me tidy it up.

X-Raym:
@jalovatt Ok, cool :)
PS : Thanks again for your contribution to this repo !

jalovatt:
Cheers.

cfillion:
@X-Raym All 44 variants of the actions are now bundled in just two packages rather than being separate (and duplicating a lot of code). That's why there is much more removed lines than added lines in this PR, but the number of actions installed is the same. 📦

X-Raym:
@cfillion
Two packages for 44 actions ? How is that possible ? ^^ I thouht it was One Script -> One Package -> One Action.

cfillion:
The relationship package:file is one-to-many. With reapack-index the file list is managed by the tags `@provides` and `@metapackage`.

So it's actually 1 package (abstract) -> X files (Y of them being actions). To be extra precise: 1 package -> X versions -> Y files (Z actions).

https://github.com/ReaTeam/ReaScripts/blob/master/MIDI%20Editor/Lokasenna_Duplicate%20selected%20notes%20diatonically.lua#L13-L14

X-Raym:
@cfillion oh yes of course, scripts can register other actions this way.

In your above example,
`Lokasenna_Duplicate selected notes diatonically/*.lua`

why `/*` ?

jalovatt:
The main script (which also serves as a library for half of the rest) is the only one that shows up in ReaPack. When you install it, that line tells ReaPack to also grab every Lua file in the subfolder.

X-Raym:
@jalovatt Understood. 👍 
That's why scripts in subfolder have : `NoIndex: true.`

One script with core code, few files for actions based on it with minimum code, brought in the action list thanks to the main package, because it is cleaner.

cfillion:
reapack-index understands a few common shell filename globbing syntaxes to select many files at once in the provides tag: https://ruby-doc.org/core/File.html#method-c-fnmatch.

PR #21, opened 2017-02-20
Moved "description" tag to "about" tag

beaunus:
As you suggested, I have moved the "description" tag to an "about" tag.  I also changed some of the header syntax.


cfillion:
Could you move the instructions segment into the about tag as well? It would look like this:

Screenshot

```
ReaScript Name: Name item takes by last marker to cut item
About:
  Names item takes according to the name of the marker that is the last marker
  to cut the items.

  Instructions:

  - Select items
  - Run the script
Author: beaunus
Version: 1.0
```

beaunus:
Thank you for your patience.  I moved the instructions to the "about" tag.  I'll use the same convention in the future.

Cheers,
Beau

PR #37, opened 2017-04-04
Update amagalma_ReaNoir track-item-take coloring utility.lua

amagalma:


Various/amagalma_ReaNoir track-item-take coloring utility.lua
```diff
@@ -31,6 +31,11 @@
 
 --[[
  * Changelog:
+ * v.2beta (2017-04-04)
```
cfillion: You might want to fix this so users can see your nice changelog on update.

cfillion:
I believe this is going to be the first script beta release shared through this repository! :+1:

Various/amagalma_ReaNoir track-item-take coloring utility.lua
amagalma: Damn! I did it again!.. Fixing! :)

amagalma:
It seems like the script does not get synchronized by ReaPack.. Is the "beta" in the version to blame?

cfillion:
The logic of synchronize packages updates works like this:

-
  - If the option "Enable pre-release globally" is enabled: install the latest available version if the currently installed version number is lower
  - Else (default behavior): install the latest available *stable* version if the currently installed version number is lower
-  Install the latest available (any) version if the currently installed version number is a pre-release

This makes pre-releases "opt-in" until the next higher stable is released and without blocking the release of more 1.X stable releases in the meantime while the beta is getting worked on.

(Pre-releases are any version number that contains at least a letter).

PR #43, opened 2017-04-10
Update amagalma_Toggle active take volume envelope visible for selected item(s).lua

amagalma:


cfillion:
Ah, I didn't notice the version bump was missing in #42!

PR #46, opened 2017-04-11
Update amagalma_Toggle take volume envelope visible for all takes of all items in project.lua

amagalma:


cfillion:
Don't forget to bump `@version 1.01`.

PR #58, opened 2017-05-11
added Nanth_item Auto Gain staging...

nofishonfriday:
added Nanth_item Auto Gain staging... scripts from [here](http://forum.cockos.com/showthread.php?t=184368) to /Items Properties.

For these scripts to work it requires "spk77_Get max peak val and pos from take_function.lua" to be put in the same folder (also added in this commit).

I think spk's script should automatically be provided when installing these, but not show up.
So I added

@provides
  [nomain] spk77_Get max peak val and pos from take_function.lua

in the scripts, and in spk's script

@noindex

Not sure if I did this correctly, someone please check. :)
Thanks.

edit:
Ok, I tried to fix the headers, but I'm stumped, sorry.

### DIFF
--- a/Items Properties/Nantho_item Auto Gain Staging - Peak vs. LUFS.lua
+++ b/Items Properties/Nantho_item Auto Gain Staging - Peak vs. LUFS.lua
@@ -1,0 +1,167 @@
+@provides
+  [nomain] spk77_Get max peak val and pos from take_function.lua
+
+--[[
+ * ReaScript Name: Item Auto-Gain Staging - Peak vs. LUFS
+ * Description: See title
+ * Instructions: Select items. Run.
+ * Screenshot: 
+ * Author: Nantho
+ * Licence: GPL v3
+ * Forum Thread: REQ: http://forum.cockos.com/showthread.php?t=182701
+ * Forum Thread URI: 
+ * REAPER: 5.0
+ * Extensions: spk77 Get max peak val and pos from take (function and example).lua
+ * Extensions: SWS extension
+ * Version: 1.0
+--]]
+
+--[[
+ * Changelog:
+ * v1.0 (2016-11-29)
+	+ Initial Release
+--]]
+ 
+-- Special thanks to X-Raym
+
+-- USER CONFIG AREA -----------------------------------------------------------
+
+threshold = -10 -- number: Default threshold to select items
+direction = "+" -- "+"/"-": Select if over or below the threshold
+
+all_items = false
+
+popup = true -- true/false: display a pop up box
+
+console = true -- true/false: display debug messages in the console
+
+------------------------------------------------------- END OF USER CONFIG AREA
+
+
+-- INCLUDES -----------------------------------------------------------
+
+local info = debug.getinfo(1,'S');
+script_path = info.source:match[[^@?(.*[\/])[^\/]-$]]
+dofile(script_path .. "/spk77_Get max peak val and pos from take_function.lua")
+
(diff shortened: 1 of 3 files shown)


cfillion:
The header is fine now, but the same filename cannot be provided by two packages (when version differs, which version of the function file should be installed without risking breaking the other scripts that use it?).

You could provide a renamed copy of the file like this, without having to make two copies of it on the GitHub repo. The new name used when installing is specified after a `>` character in the provides line next to the original path:

```
Provides: [nomain] spk77_Get max peak val and pos from take_function.lua > Nantho_Item Auto-Gain Staging - Peak vs. RMS.lua/spk77_Get max peak val and pos from take_function.lua
```

The scripts's header could also use a nice `About` section and `Link` rather than the unused Instructions/Forum Thread/Extensions/License/etc.

nofishonfriday:
Thanks, phew build passed this time. 👯‍♂️ 
Hope it's fine now.

Re: header, this was done by the original script author this way so I refrained from modifying it originally.
But you're right, it makes sense adding these sections for better ReaPack useabilty. Done.

Items Properties/Nantho_item Auto Gain Staging - Peak vs. LUFS.lua
```diff
@@ -0,0 +1,169 @@
...
+console = true -- true/false: display debug messages in the console
+
+------------------------------------------------------- END OF USER CONFIG AREA
+
+
+-- INCLUDES -----------------------------------------------------------
+
+local info = debug.getinfo(1,'S');
+script_path = info.source:match[[^@?(.*[\/])[^\/]-$]]
+dofile(script_path .. "/spk77_Get max peak val and pos from take_function.lua")
```
cfillion, line 47: Shouldn't this be changed to use the new path to the file?

Items Properties/Nantho_item Auto Gain Staging - Peak vs. LUFS.lua
```diff
@@ -0,0 +1,169 @@
+--[[
+ * Provides: [nomain] spk77_Get max peak val and pos from take_function.lua > Nantho_Item Auto-Gain Staging - Peak vs. RMS.lua/spk77_Get max peak val and pos from take_function_Copy2.lua
```
cfillion, line 2: This could go in a "Peak vs. LUFS" folder to match the script's name, rather than both Copy1+Copy2 in one "RMS" directory (and then suffixing with CopyX is no longer necessary).

nofishonfriday:
D'oh, you're right.
I'm closing this PR and open a new one when I tested in my own ReaPack repo it works properly.

nofishonfriday:
Hm..trying in my Reapack repo, the function script shows up in ReaPack browser despite the @noindex tag.
http://i.imgur.com/5tbPLaO.jpg

I'm out of wisdom.
Sorry to say, but I give up, this procedure is over my head (please don't take it as meant negative for ReaPack).

cfillion:
Sorry I didn't catch this before:

```
--[[
 * noindex
 * Version: 1.0
--]]
```

`noindex` like this alone isn't read as a tag by the indexer, but as a comment it should ignore.

To make it a tag it must have a value: `NoIndex: true`.  
The @-syntax doesn't require a value – it assumes true if it's missing, so `@noindex` works too.

EDIT: Files having noindex enabled don't need to have a version tag since they're skipped altogether.

nofishonfriday:
Thanks for bearing with me. I'll try again. :)

PR #60, opened 2017-05-12
#59 fix script_path

nofishonfriday:
uh sorry, only checked the other one previously and just noticed now


cfillion:
I've edited the version number to be safe in case someone already installed the broken one. 👍

nofishonfriday:
Thanks.
(Quite a bit of learning experience for me. :))

PR #65, opened 2017-08-02
Fix "u/0005 appears in all markers" in beaunus_Add markers for lyrics in selected items.lua

beaunus:
Prior to this fix, the script was adding an annoying control character "u/0005" at the beginning of all the marker names.  

It was mostly inconsequential, but if the marker names were used to define file names, the file names would inherit the annoying control character.


cfillion (changes requested):
Please bump the version tag at line 13 to 1.0.1 as well.

The wordpress-style changelog syntax alone does not set the package's version.

beaunus:
I'm sorry about that.  Thank you for the reminder.

cfillion (approved):

PR #66, opened 2017-09-03
Clip Splicer

beaunus:
I would like to contribute a sizeable script to your repository.

I've tried my best to conform to your documentation standards.  Please let me know if anything needs adjustment or explanation.

Also, please forgive me if I made any mistakes merging your work into my branch.  GitHub says I'm ~1400 commits ahead of your branch.  I'm still getting used to merging branches.

Thank you.
Beau


cfillion:
Very nice documentation!

I assume you will want the files in schema/ and examples/ to be installed along with your script. Additional files must be specified in a Provides tag like this:

```
Provides:
  schema/*.json
  examples/*.json
```

Your script will look like this in the package browser:



It might be a good idea to move the .py file directly into the Various directory so that the category becomes just "Various". Do you intend Clip Splicer to be a multi-scripts package in the future, or will it be just the one import JSON file?

About the "1400 commits ahead": don't worry about it for PR, I will copy the files manually. This is happening because your fork is very out of sync with our master branch. For next time please synchronize your fork with the upstream repo (or just delete/recreate it on GitHub – same result) before pushing commits. Commiting to new branches instead of `master` is a good way to avoid many of these problems.

By the way, this will be the first python script of this repository! 🎆

PR #71, opened 2017-11-26
Create ED_Select items with take FX.lua

mkruselj:


cfillion:
ReaScript Name and Description are the [same tag](https://github.com/cfillion/reapack-index/wiki/Packaging-Documentation#description). You probably want to use [About](https://github.com/cfillion/reapack-index/wiki/Packaging-Documentation#about) for the documentation to be displayed as such instead.

```lua
-- ReaScript Name: Select items with take FX
-- Version: 1.0
-- Author: EvilDragon
-- About:
--   Fairly self-explanatory - this script will select all items that have take FX loaded in any take - active or not.
--
--   Requires REAPER 5.0+ and SWS. Licensed under GPL v3.
```

or

```lua
-- @description Select items with take FX
-- @version: 1.0
-- @author EvilDragon
-- @about
--   Fairly self-explanatory - this script will select all items that have take FX loaded in any take - active or not.
--
--   Requires REAPER 5.0+ and SWS. Licensed under GPL v3.
```

PR #78, opened 2017-11-27
Tons of script naming fixes

mkruselj:


X-Raym:
@cfillion What do you think about this pull request ? It seems to only affect action description in Reapack browser so I don't think it breaks anything else. But the PR do have conflict now.

@mkruselj you were very motivated to correct all that ! ✌️

mkruselj:
I was indeed quite motivated. :D

X-Raym:
@cfillion what do we do about this ? Do you think it is relevant to fix these ?

X-Raym:
@mkruselj Hi ! This is a very good PR, we should have accepted since a long time, but the repo has change quite a bit since then, and it is too difficult to merge directly. @cfillion will try to answer this pro grammatically so it also works for newly added scripts. He will take care of description tag so that it matches file names. If they are more subtle changes, then it may be lost, but it would requires too much time to invest. We also count on original scripts authors to fix their typos 😉

I close this cause it will be handle differently.

This was the last opened PR on any ReaTeam repo 😎

mkruselj:
Sounds good!

PR #80, opened 2018-01-10
Added Save/Load Region Render Matrix scripts

ausbaxter:
Initial commit.
Scripts use project ext state to save matrix settings to project.


ausbaxter:
Source Thread:
https://forum.cockos.com/showthread.php?p=1938225#post1938225

Regions/Ausbaxter_Load region render matrix slot 1.lua
```diff
@@ -0,0 +1,128 @@
+--[[
+ * ReaScript Name: Load region render matrix slot n
```
cfillion: Please use a different name for each package. Currently they all have the same "slot n". But read below first: there is a better way to package this without creating one package for every slot. (One ReaPack package can contain multiple scripts and files)

Regions/Ausbaxter_Load region render matrix slot 1.lua
```diff
@@ -0,0 +1,128 @@
...
+ * Description: Allows quick storage and recall of reaper render matrix.
+ * Instructions: Make track selections in region matrix, use the "complementary Save region render matrix slot n" to save the setup. 
+      Use the 'load' scripts to recall.
```
cfillion: You seem to be using a header style from old scripts. While reapack-index does understand some of these tags, most are discarded or won't behave as intended.

Please refer to the [packaging documentation](https://github.com/cfillion/reapack-index/wiki/Packaging-Documentation) and [examples](https://github.com/cfillion/reapack-index/wiki/Examples) pages for proper headers.

- ReaScript Name becomes Description
- Description and Instructions becomes About
- Author URI, Repository URL, File URI etc becomes Links
- Changelog syntax is different (no longer a separate block)

Regions/Ausbaxter_Load region render matrix slot 1.lua
```diff
@@ -0,0 +1,128 @@
...
+ 
+--[[
+ * Changelog:
+ * v1.0 (2018-01-09)
+  + Initial Release
+--]]
+--------------------User Area---------------------------------------------------------------------------
+--To create more save slots duplicate this script and change the slot number variable
+
+slot_number = 1 --only use integers
```
cfillion: It would be better to bundle all these slots and actions and into a single installable package. I think there is no point in allowing users to install only "save slot 4" without the others actions.

By making a single package you will be able to reuse the same code for each slot without copy-pasting on your end. It guarantees that all slots are installed and updated together.

See [Multiple slots](https://github.com/cfillion/reapack-index/wiki/Examples#multiple-slots) in the Examples page.

Using this technique you can have down to just one single file to maintain and upload here.

cfillion (changes requested):
Thanks for your first contribution to this repository!

I have a few requests to improve packaging. Feel free to ask questions if you have any.

PR #81, opened 2018-01-10
Update FnA_Ensure toggle OFF (item grouping).lua

fnaFnaFnAFNA:
Added comment. User tried to put in MIDI Editor section.


cfillion:
Merged!

If you wanted to put this script in both sections, you can do it like this:

1. In the reapack-index header: `Provides: [main=main,midi_editor] .` (<-- there's a trailing dot!)
2. And in the script itself you can detect from which section it was ran from using `reaper.get_action_context`

fnaFnaFnAFNA:
I don't really know what to use for a midi example...

The user in the discussion thread had a pretty unique thing in mind.

PR #83, opened 2018-01-10
Create FnA_Join Only Selected Notes With Spacing Below Threshold - In…

fnaFnaFnAFNA:
…put and No Input Versions.lua

Not sure about provides [main=midi editor] 
Other header stuff...


cfillion:
Please remove the redundant `.lua` from the description tag. There is also very few scripts and native actions using Title Case. Perhaps "Join selected overlapping and consecutive notes" would make it easier to find in the action list if the ReaPack name and both filenames contain this?

The provides tag is correct. Thank you for bundling both similar actions in a single package! 👍

(Note that simply putting a script in the MIDI Editor folder implicitly sets the default Action List section accordingly when using `[main]`)

fnaFnaFnAFNA:
Hope that did it...

PR #84, opened 2018-01-11
Update FnA_Join selected overlapping and consecutive notes with spaci…

fnaFnaFnAFNA:
…ng below threshold - input and no input versions.lua


fnaFnaFnAFNA:
Fix lines related to choosing whether to use input box version or not. About section improvement.

PR #86, opened 2018-01-12
Added Render Item Columns

ausbaxter:
GUI interface to render selected items as columns.


ausbaxter:
hmm, seems all my previous commits are being requested? Not sure how to prevent that from happening. :/ Sorry about that

Various/ausbaxter_Render item columns.lua
```diff
@@ -0,0 +1,3878 @@
+--@description ausbaxter_Render item columns
```
cfillion: Please don't repeat the author name in the package name. This clutters the list and breaks alphabetical sorting.

cfillion, line 1: Please don't repeat the author name in the package display name (description tag). It clutters the package list and breaks alphabetical sorting (it's OK in the file name).

(Packages doing this currently will be edited in the future)

Various/ausbaxter_Render item columns.lua
```diff
@@ -0,0 +1,3878 @@
...
+--@version 1.0
+--@author ausbaxter
+--@about
+--    # Render Item Columns
+--
+--    This package provides a gui interface to render columns of selected items. Has optional tail length, rename capabilities, and its also possible to add regions based on item columns.
+--    GUI created using Lokasenna's GUI 2.0. https://forum.cockos.com/showthread.php?t=177772
+--@provides
+--    ausbaxter_Render item columns.lua
```
cfillion, line 10: This provides tag is not necessary in this case because the current script file is implicitly added to the package's file list (however it doesn't hurt to explicitly list it like this).

cfillion (approved):
I released ausbaxter_Render item columns.lua manually (removing the redundant "ausbaxter_" from the package display name – see comment below – 2ea50cd32ae61da29c4589ffc7e2e92375ab6470).

You can prevent git conflicts and repeated commits by making all changes in branches (one per PR) instead of directly to master.

For a quick-and-dirty solution you can delete your fork and re-create it to reset your master branch.

I see you have made changes to ausbaxter_Save and Load region render matrix.lua as well: increase the version number (eg. 1.0.1) and optionally edit the changelog if you want existing users to get the update (otherwise only new users will get the modified code).

ausbaxter:
Thanks for your help and info cfillion!

PR #91, opened 2018-02-07
Update README.md

bfut:
forgot to update readme concurrent to last week's contribution


bfut:
looks like another pull request was committed like two minutes earlier. be back in a few minutes

PR #94, opened 2018-03-27
Create mschnell_Mute Track X.eel

mschnell1:
Is creating a directory "Utility" in scripts OK ? 
-Michael


mschnell1:
How to correctly add an "about:"  MarkDown description to an eel ReaScript file ?
-Michael

cfillion:
Shouldn't this go into the "Tracks Properties" folder? The syntax(s) for the `about` is the same for EEL, but you'll need to make it a valid comment so REAPER doesn't complain.

Perhaps the name could include that this script is MIDI CC/OSC only to further differentiate it from the native "Toggle track mute" and "Track: Set mute for track n (MIDI CC/OSC only)"?

If you want you can make a package containing both Mute and Unmute actions using the [`@provides`](https://github.com/cfillion/reapack-index/wiki/Packaging-Documentation#provides) tag to install multiple files. In the example below I make it install the same file under two different names. Ignore this (don't use `@provides`) if you want the Mute and Unmute actions to be installable or updatable separately.

```c
// @description Mute or unmute track X (MIDI CC/OSC only)
// @version 1.0
// @author Michael Schnell
// @about
//   # Your Package Title
//
//   Longer description/documentation for this package in *markdown*.
// @changelog Initial release
// @metapackage
// @provides
//   [main] . > mschnell_Mute track X (MIDI CC/OSC only).eel
//   [main] . > mschnell_Unmute track X (MIDI CC/OSC only).eel

mute = extractFromFileName()
ShowMessageBox("Hello World", "Script code here", 0)
```

You could also use multiline `/* */` comments if you prefer. reapack-index doesn't care about the language-specific comment syntax as long as it remains consistent across all header lines.

mschnell1:
Sounds good. 
Thanks !
trying ....
... But what is "mute = extractFromFileName()" ?
And what is "MIDI CC/OSC only"
-Michael

mschnell1:
And again: 
"files changed" shows the unmerged text. 
"view" opens a kind of Editor
Pencil symbol -> error message: "You  must be on a branch ..."

-Michael

cfillion:
The filename has an extra space and colon at the end: `mschnell_Mute Track X.eel :`.

> ... But what is "mute = extractFromFileName()" ?

Currently you have two mute/unmute scripts with `mute = 1;` and `mute = 0;` respectively. If you use the [`@provides`](https://github.com/cfillion/reapack-index/wiki/Packaging-Documentation#provides) tag in the way I used above, you will have only one codebase installed under two filenames.

So in this case the value of `mute` needs to be derived from the filename. You already have access to the filename from your `get_action_context` call.

> And what is "MIDI CC/OSC only"

REAPER puts this in parenthesis at the end of the built-in actions that can't be used from the action list.

> And again:
> "files changed" shows the unmerged text.
> "view" opens a kind of Editor
> Pencil symbol -> error message: "You must be on a branch ..."

Normally when you click on the "File changes" tab the correct pencil-shaped button is at the right of the "View" button and the "Open in GitHub Desktop" button. But it's hidden when the Pull Request is closed. I'm reopening this PR so the button reappears.

mschnell1:
This is the very first time I get in contact with ReaScripts, so I need to do more investigating regarding the file name detection....
-Michael

mschnell1:
The "create two files with the same content" is a really cool feature of ReaPack's for exactly this purpose !

Thanks,
-Michael

cfillion:
The filename ends with ` :` instead of just .eel. Windows (and REAPER) won't like that! Also please move it to `Tracks Properties` (or at least `Various`, which is the current equivalent of JSFX's catch-all `Utility`).

mschnell1:
Retrying ....

mschnell1:
OK. Seems I succeeded in renaming the file appropriately. 

Is there any chance to define a kind of common "Category" this and the "Live Utilities" I created in the JSFX directory could reside in together ? 

BTW.: Why don't I get e-mail notification of the proceedings in this (and several other) "Projects"  ? (While  I do get those with even other projects.)

cfillion:
Various it is then. Seems all good now, merging! 😺

cfillion:
Oops I did not see your last comment.

"Live Utilities" sounds like a good idea to me. Definitely better than the uninformative "Various"! Just know that moving packages to a different category will break* existing projects/actions using them (moving is essentially deleting + creating a new one).

Categorization does need an overhaul for documented & consistent naming across all repositories (I'll only have time to work on this once my own live concert is over.)

> BTW.: Why don't I get e-mail notification of the proceedings in this (and several other) "Projects" ? (While I do get those with even other projects.)

What does it say in the sidebar (top right) of this page, under "Notifications"? Mine says this <https://i.imgur.com/03RwbUr.png> and I get one email for every action you do on this Pull Request.

(*Actually ReaPack would just mark the old one as Obsolete and ask users whether they want to uninstall it. So projects only break if users say "yes, uninstall".)

mschnell1:
I don't supposed that these Utilities are widely used, yet, so I think it's a good time to move them right now. Would I be able to create a new directory or would you install it (at least in in "/Effects" and in "/Scripts") ? 

Do you intend to create a new tag "category:" (or similar) or do you want to stay with the subdirectories as a kind of category ? 

The sidebar also says : "You’re receiving notifications because you modified the open/close state"

BTW. This project here says "Pull request successfully merged and closed", but I don't see it in the ReaScript Browser after  synchronizing. 

-Michael

mschnell1:
> The sidebar also says : "You’re receiving notifications because you modified the open/close state"

Nonetheless, on the top I could set the Watch option apparently  for "ReaTeam/ReaScripts". Maybe this is necessary ...

-Michael

mschnell1:
Would it make sense to have a directory "Live Utilities" a swell in "Scripts", as in "effects" (and maybe later in other locations as well ? 

-Michael

mschnell1:
I finally found the package and installed it. 

Is it really intended that there are two directories in "various" (called "mschnell_Mute track X (MIDI CC" and "mschnell_UnMute track X (MIDI CC") both containing a single file called "OSC only).eel" ????

Obviously this is due to the "main . " containing a "/" slash. 

As I am not familiar with this syntax, I can't fix it :( 

Again, I fail to find how to create a pull request for this file to allow me to edit it :(

-Michael

mschnell1:
Seemingly when trying to edit the file I have created a new file or project. 

I hope I did not create a great mess.... Should this "branch" be deleted ? 

(BTW.: you seem to have merged the new file but I did not get any notification.) 

Now trying to find the new file ...
-> #96 

-Michael

mschnell1:
Now while it works...

>REAPER puts this in parenthesis at the end of the built-in actions **that can't be used from the action list**.

I still don't understand that part.

-Michael

PR #97, opened 2018-04-09
Create JerContact_Open Project Based on File in Clipboard.lua

JerContact:
I love the idea of ReaPack!  This is my first contribution to the ReaPack community, and I think it's a great resource for sound people in games!  I'm excited to share more soon!


cfillion:
Thanks for the contribution! Before I can merge this, please edit the following:

- "ReaScript Name" is an alias for the [`@description`](https://github.com/cfillion/reapack-index/wiki/Packaging-Documentation#description) tag. Using both is redundant (same as having two `@description` – the second one overwrites the first).
- Please use sentence case in the description so it better matches existing REAPER actions and most other scripts
- Move the script in one of the subfolders to set the package's category (file at the root are not be indexed by reapack-index). "Project Properties" or "Various" seems like good candidates.

PR #98, opened 2018-04-09
Create JerContact_Open project based on file in clipboard.lua

JerContact:
I couldn't figure out how to move the script from the main ReaScripts folder, so I just copied the updated code per your specs into the proper folder.  Let me know if that works!


cfillion:
Description and category is good now but the ".lua" extension is missing from the filename. You can rename (or move, it's the same) from the edit page (filename field at the top).

PR #99, opened 2018-04-10
Update to account for MAC or PC

JerContact:
updating script to work on both platforms, also updating for windows when you right click and you get added quotations for the pathname


cfillion:
Please bump the version to a higher number if you wish the update to reach existing users (+ changelog). Otherwise the update will only be applied to new installs or manual reinstalls (if this is what you want I'll merge this PR as-is).

PR #100, opened 2018-04-12
fixed bug - file couldn't be overwritten

JerContact:
pcm_source call in my script kept file "in use" and you would need to restart reaper to render out the file of the same name, fixed this!


cfillion:
There's no need to repeat old changelogs in the tag. Previous changelogs are preserved and can be read at any time from the package's About window. When updating, users are presented with all changelogs after the version they previously had installed.

JerContact:
updating changelog

PR #112, opened 2018-06-26
Lokasenna Lua GUI Framework: Reapack compatibility

jalovatt:
Yay, I'm finally getting around to a ReaPack release so people don't have to distribute the library themselves.

@cfillion, would you mind having a look to make sure this will install the way I'm expecting? There are two metapackages in the Development folder, each of which is intended to provide their respective subfolders, maintaining the folder structure on the user's drive (i.e. _Reaper/Scripts/ReaTeam/Development/Lokasenna_GUI/Library_)


jalovatt:
I see that the build failed. Adding NoIndex tags as needed...

jalovatt:
Not sure how to resolve the Duplicate File errors I'm getting. :/

cfillion:
Some files were matched by both `*.*` and `**/*.*` (`**` enables recursive search) causing them to be listed twice.

I also renamed the packages to `Lokasenna_GUI library v2.lua` and `Lokasenna_GUI library v2 (developer tools).lua` (both internal package ID and displayed name, matching the thread title). I can revert this if you prefer the old name (I think "package" shouldn't be in the package ID/name though).

Lokasenna's GUI library for Lua

https://cfillion.ca/files/.lokasenna-test.xml

jalovatt:
That looks fine.

I did some Googling about Ruby's file-matching but couldn't figure out the proper wildcard terms. Thanks.

PR #113, opened 2018-07-07
Create solger_ReaLauncher.lua

reasolger:
Hi, 

This is the first release version of the ReaLauncher for public beta testing.
The discussion thread can be found here: https://forum.cockos.com/showthread.php?t=208697

### DIFF
--- a/Project Properties/solger_ReaLauncher.lua
+++ b/Project Properties/solger_ReaLauncher.lua
@@ -1,0 +1,659 @@
+-- @description ReaLauncher: A custom (recent) projects and project/track-templates launcher 
+-- @author solger
+-- @version 0.1
+-- @changelog
+--   + Initial release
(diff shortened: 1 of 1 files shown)


Project Properties/solger_ReaLauncher.lua
```diff
@@ -0,0 +1,659 @@
+-- @description ReaLauncher: A custom (recent) projects and project/track-templates launcher 
+-- @author solger
+-- @version 0.1
+-- @changelog
+--   + Initial release
+-- @screenshot https://forum.cockos.com/attachment.php?attachmentid=34088&stc=1&d=1530957174
+-- @about
```
cfillion: The entire content of a tag must be indented the same way when spanning over multiple lines. Here, the `about` tag would only pick up the first two lines below (which are indented with two spaces relative to the tag). The rest of the description must also be intended the same way to be considered as part of the `@about` tag.

reasolger: Thanks for the info.

PR #115, opened 2018-07-08
ReaLauncher update

reasolger:
update with minor bugfixes


reasolger:
Sorry for any inconveniences caused at the beginning.
I'm still trying to figure things out about the best way for doing script updates ...

cfillion:
Editing the file from the edit button on ReaTeam/ReaScripts directly instead of from the fork (reasolger/ReaScripts) will prevent any risk of conflicts (and also does one different branch per PR which is good practice).

To keep the history cleaner, all commits from PRs are squashed into a single one when merged. This has the side effect that the commits merged here ends up being different from the one on the fork. This was the cause of the git conflict: the PR was based on the original commits and not the one actually merged.

PR #134, opened 2018-12-31
Release tompad_MusicMath - bpm/tone to ms/Hz converter v1.1

tompad2:
Added coverter note to Hz
Added tooltip "Click to copy to clipboard"
Added flashing green when clicking on ms or Hz label


cfillion (changes requested):
The filename is different from the previous version (was tompad_MusicMath.lua, now tompad_musicmath.lua). This creates a second package. From reapack.com/upload, you can open the previous version's data by clicking on it when it shows up while typing the display name (instead of restarting from scratch every time).

Also the display name (`@description`) should not contain the author prefix ("tompad_"): that's what the author field is for.

PR #148, opened 2019-02-03
Release Disable multichannel metering on selected tracks v1.0

Edgemeal:


cfillion:
Please upload the script's code directly in the file (paste below the auto-generated metadata header, or click on "Load local file"). In this case you don't need to add a second file.

Externally-sourced files should only be used for large binary files (themes, extensions) and the URL should be unique to the package's version.

PR #152, opened 2019-02-04
Release jm_scripts v1.0

jmieszkowski:


cfillion (changes requested):
Please upload the script's code directly in the file (paste below the auto-generated metadata header, or click on "Load local file") instead of using an externally-sourced URL. It also appears https://github.com/jmieszkowski/ReaScripts/ does not exist anymore.

Track color is a property of tracks, so the category should be set to Track Properties.

The display name should also be set to properly describe the script being uploaded ("jm_scripts" doesn't). I suggest something along "Set the color of selected tracks from their parent tracks" (if this is what the script does).

PR #154, opened 2019-02-06
Release Set first selected point to last selected point value in AI v1.0

Edgemeal:


cfillion:
Given these two scripts (first to last & last to first) are nearly identical, would you like to bundle them into a single package? This would guarantee all users have both installed at the same version and make updating them easier.

You can create such a bundle like this (this way will make both actions use the same source code):

1. Setting the main package file's resource type to "Don't install" (so that is stays only on the repository)
2. Adding the two actions as new "additional files" whose file source are set to the package file and Action List set to Main
   - edgemeal_Set first selected point to last selected point value in AI.lua
   - edgemeal_Set last selected point to first selected point value in AI.lua

```lua
local scriptName = ({reaper.get_action_context()})[2]:match("([^/\\_]+)%.lua$")

if scriptName:match('^Set first') then
  reaper.SetEnvelopePointEx(env, a1, p1, _, v2, _, _, true, false)
else
  reaper.SetEnvelopePointEx(env, a2, p2, _, v1, _, _, true, false)
end

reaper.Undo_EndBlock(scriptName, -1)
```

If you prefer to instead keep these two actions separately installable as individual packages, I'll merge your two pull requests as-is.

Edgemeal:
Ya just leave them as two separate actions, Thanks!

PR #169, opened 2019-03-15
Release Toggle show monitoring FX chain v1.02

Edgemeal:
Fix for MacOS (Not closing chain window)


cfillion:
Why changing the category from FX to Various? (Changing category is the same as making a new package and deleting the old one, making the later "obsolete" and prompt for uninstalling.)

Edgemeal:
> Why changing the category from FX to Various? (Changing category is the same as making a new package and deleting the old one, making the later "obsolete" and prompt for uninstalling.)

Well its not FX. Is It? I'd remove v1.0  but don't know how, Sorry, this github/repack stuff is too complicated, I'll just post to forum! ;-)

cfillion:
It deals with an FX chain, so in my opinion FX-related is closer to describing its usefulness than "it does various things".

(As for deleting, I meant that changing the category does it automatically for you.)

PR #170, opened 2019-03-15
Create Archie_Add Repository Archie-ReaScripts.lua

ArchieScript:


cfillion:
I'd rather not have all repositories put "add me" scripts here. We've had a "import everything" script for while but it has been deleted as it was obsolete and incomplete. Now we can copy/paste from the website to import all repositories at once and there's a direct link to the repository list page in the Import window. The "Repositories" page on the website is quite popular (around 50% more than the user guide!).

ArchieScript:
Everything is clear, Thank you! It is a pity of course that you can not add((.
But as for me, a very large number of users (especially beginners) can not figure out how to install "reapack" not to mention some links that need to be added separately and individual links are used only by advanced users of "Reaper", and beginners do not give a choice. 
It seems to me that with the installation of "repack" all repositories should be added to "Reaper" at once, but with the synchronization disabled, and the user would have chosen what to add and what not to add.
But even if there is no possibility or desire to add all the repositories in the "Reaper" at once, then why not add the scripts (on desire who wants) "Add me" to the repository, which is installed by default, a beginner would see it and download it (But the script should ask whether to install this repository or not, and are not arrogant to add), if he is looking for something and can not find, and so he is about these repositories(links) and does not know, and know only forum users.
Incorrect all the same logic to add scripts.
And yet it's a pity that you can't add, but still Thank you for everything, first of course for myself "reapack"

PR #174, opened 2019-04-07
Create amagalma_Create Impulse Response (IR) of the FX chain of the s…

amagalma:
…elected track.lua


cfillion:
I don't know if you've seen it, but we have a new way to upload scripts to ReaPack now: https://reapack.com/upload/reascript.

PR #185, opened 2019-05-10
Release Window for live use with 6 big customizable buttons and showing region names, colors and progress bar v1.0

stepanhlavsa:


Regions/stepanhlavsa_Window for live use with 6 big customizable buttons and showing region names, colors and progress bar.lua
```diff
@@ -0,0 +1,270 @@
+-- @description Window for live use with 6 big customizable buttons and showing region names, colors and progress bar
+-- @author Stepan Hlavsa
+-- @version 1.0
+-- @provides [windows] .
```
cfillion: Is there a reason for this to be available exclusively for Windows? From what I can tell it should also work on macOS and Linux.

Also "Window for live use with 6 big customizable buttons and showing region names, colors and progress bar" is a pretty long and verbose name. Perhaps it could be shortened a bit to fit better with the way other REAPER actions are named. What would you think about something shorter such as "Big region progress bar with 6 assignable buttons"?

stepanhlavsa:
Hi,

it will probably work but I can not test it on Linux or Mac, therefore I can not be sure.

 

You may use short name, like „Live play window with progress bar“ or the name you suggested.

Thank you,

Stepan Hlavsa

PR #192, opened 2019-06-04
Release obliques v1.0

tompad2:


Various/tompad_obliques.lua
```diff
@@ -0,0 +1,137 @@
...
+--
+--   tompad_obliques is a reascript that randomly shows text
+--   from Oblique Strategies by Brian Eno and Peter Schmidt.
+--
+--   Any comments on coding, requests, bugs etc is welcome! PM me (tompad) on Reaper Forum (https://forum.cockos.com/member.php?u=19103).
+--
+--   Credits
+--   Oblique Strategies by Brian Eno and Peter Schmidt
+
+-- @description tompad_obliques - randomly shows text from Oblique Strategies
```
cfillion: There are two metadata headers in this main file. Only the first one will be used.

Various/tompad_obliques/obliques.lua
```diff
@@ -0,0 +1,215 @@
+-- @noindex
```
cfillion: Loading the random messages from this obliques.lua will likely choose the first two metadata lines sometimes (`@noindex` is required due to the .lua file extension–a file without an extension would not need any metadata).

tompad2: Ok. But what resource type should I select in Upload Tools for that file? I had a .txt file and tested different types but the path to .txt changed to a one I didn't want. Can I edit file here online or will I need to do a new upload from Upload Tools?

cfillion: The resource type should remain the same (ReaScript) to have the file installed into the Scripts folder. The file extension can be set in "storage file name" independently of the resource type (I suggest not using any extension for that file–.txt is used in REAPER and ReaPack for MIDI note/CC name files).

You can edit the files right here from this pull request's page on GitHub.

```lua
local file = path_to_script .. "obliques.lua"
```

...should be


```lua
local file = path_to_script .. "tompad_obliques/obliques.lua"
```

tompad2: Ok, I made a change - what happens now? (Yes I'm a noob on GitHub;-))

tompad2: And another question......why dont my 2 files live in one folder? Now it looks like my obliques-file is in a folder but the script is not in same folder...

cfillion: Now the script can safely be published (I'll merge this tomorrow, in case you want to move the obliques file before the script gets released).

The upload tool automatically puts additional files into a subdirectory to keep the repository tidy when scripts consist of many files. (This can be bypassed by prefixing the storage file name with `../`.)

tompad2: Ok, so if I am satisfied with having obliques file in a subdirectory there is no more I need to do?
After some thinking I think its good to have it in subdirectory, as you say, keeps repository tidy.
Looking forward to see my script in ReaPack - thanks for all cfillion!

PR #213, opened 2019-08-31
Update solger_ReaLauncher.lua

reasolger:
+ [Project Templates]: Code update for using the new 'project template loading' API function prefix introduced with Reaper v5.983
+ Audio Preview: The volume knob now indicates if a preview file is available for the selected list entry by changing its color to silver
+ General: Improved code for reading files from folders - no more command window pop ups :)
+ General: Added first implementation for subfolder listing and selection
+ General: Automatic load and refresh of list entries is now only done for the focused tab (instead of loading the lists for all tabs simultaneously)
+ Key inputs: Adjustment for list scrolling behavior

### DIFF
--- a/Project Properties/solger_ReaLauncher.lua
+++ b/Project Properties/solger_ReaLauncher.lua
@@ -1,8 +1,14 @@
 -- @description ReaLauncher
 -- @author solger
--- @version 1.5.1
+-- @version 1.6
 -- @changelog
---   + Bugfix: New Project button not working
+--   + [Project Templates]: Code update for using the new 'project template loading' API function prefix introduced with Reaper v5.983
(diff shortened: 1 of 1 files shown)


reasolger:
Hi @cfillion.

This time I edited/updated the code again directly since the ReaPack Upload tool is always throwing a 'Could not parse header of Project Properties/solger_ReaLauncher.lua. Server replied with 500 Internal Server Error" when trying to load the package for editing. (I tried multiple times with several hours inbetween). This error didn't show up the last time when using the Upload tool.

cfillion:
Thanks for the report! That bug was only triggered when reading large files. I've deployed a fix.

PR #224, opened 2019-10-13
Release Smart split MIDI item (trim left note parts at cursor) v1.0

reasolger:
First version

### DIFF
--- a/Items Editing/solger_Smart split MIDI item (trim left note parts at cursor).lua
+++ b/Items Editing/solger_Smart split MIDI item (trim left note parts at cursor).lua
@@ -1,0 +1,82 @@
+-- @description Smart split MIDI item (trim left note parts at cursor)
+-- @author solger
+-- @version 1.0
+-- @changelog First version
+-- @about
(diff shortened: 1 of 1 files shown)


cfillion:
There's a lot of identical code in these three scripts. I think it might be a good idea to bundle them into a single "Smart split MIDI item" package (one package, three installed actions).

One way to do this is with a single codebase detecting from which action it is being run by reading its own filename: https://github.com/cfillion/reapack-index/wiki/Examples#multiple-slots.

On https://reapack.com/upload/reascript, set the resource type of the primary source file (solger_Smart split MIDI item.lua) to "Don't install". Then click on its "Duplicate file" button three times to create the individual actions (set "Install file name" to the desired action filename for each three).

(Of course, if you prefer to keep these available as three individual packages, I'll merge your pending pull requests as-is.)

reasolger:
Ah, OK. Thanks for the info. I'll look into this.

reasolger:
I'll close the pull requests for now. Will upload a new (bundled) version in the next days.

PR #229, opened 2019-10-14
Release Smart split MIDI item (bundle) v1.0

reasolger:
First version

### DIFF
--- a/Items Editing/solger_Smart split MIDI item (bundle)/solger_Smart split MIDI item (bundle).lua
+++ b/Items Editing/solger_Smart split MIDI item (bundle)/solger_Smart split MIDI item (bundle).lua
@@ -7,7 +7,7 @@
 --   [main] . > solger_Smart split MIDI item 1 (trim shorter note parts at cursor).lua
 --   [main] . > solger_Smart split MIDI item 2 (trim left note parts at cursor).lua
 --   [main] . > solger_Smart split MIDI item 3 (trim right note parts at cursor).lua
---   [nomain] solger_Smart split MIDI item (bundle)/solger_Smart split MIDI item functions.lua
+--   [nomain] solger_Smart split MIDI item functions.lua
 -- @about
 --   These scripts split a MIDI item and trim the (shorter, left or right) note parts at the cursor position. The threshold of the note part length which should be trimmed can be adjusted in each script (by default all notes under the cursor are trimmed).
 --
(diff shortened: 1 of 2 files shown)


reasolger:
Hi Christian,

first attempt to check in a bundled package :)

I have two questions:

1) The _solger_Smart split MIDI item functions.lua_ file should not be shown in the Action List (it should only be installed). According to the Packaging documentation, the [nomain] option is used for that (default for additional files). But I couldn't find an option to set or add the [nomain] option to the provide section via the Upload tool (unless I missed it somehow).

Is this something which currently requires editing the file directly in GitHub (which I did this time)?
Or is having no option parameters in front of a file entry the same as having [nomain] in front?

EDIT:

2) All installed lua files should preferably be located in the 'Items Editing/solger_Smart split MIDI item (bundle)' subfolder. Is the last version of the entries correct in this case?
> -- [main] . > solger_Smart split MIDI item 1 (trim shorter note parts at cursor).lua
-- [main] . > solger_Smart split MIDI item 2 (trim left note parts at cursor).lua
-- [main] . > solger_Smart split MIDI item 3 (trim right note parts at cursor).lua
-- [nomain] solger_Smart split MIDI item functions.lua

Thanks

cfillion:
1. Yes, `[nomain]` is the default value for additional files.
2. The package file itself should not be in a subfolder because this creates a new category. To use the Items Editing category the package file should be directly in that folder and only have the additional hosted files in the subfolder.

Alternatively everything can be done with a single Lua source code in this case (no need for an additional hosted functions file). I adapted the script to do so here: #231. Using the following snippet to detect the trim mode:

```lua
local scriptName = ({reaper.get_action_context()})[2]:match("([^/\\_]+)%.lua$")
local trimMode = ({
  shorter = 1,
  left    = 2,
  right   = 3,
})[scriptName:match('%(trim (.+) note parts at cursor%)$')]
assert(trimMode, 'invalid filename')
```

Uploaded like this:
Screenshot

Let me know whether you prefer the functions file or single source file approach, I'll merge either this PR (after moving the package file to the Items Editing category) or #231.

reasolger:
> Let me know whether you prefer the functions file or single source file approach, I'll merge either this PR (after moving the package file to the Items Editing category) or #231.

Cool. Yeah, I prefer the single file approach. So let's go with your updated version #231.

Thanks for all your help and info!

PR #244, opened 2019-10-30
Updated two region scripts

Landeplage:
Small bugfixes.


cfillion:
I have removed the duplicate `@changelog` tag: the changelogs of all previous versions are stored in the repository index and does not need to be kept in the script.

(By the way, https://reapack.com/upload is the new recommended way to submit scripts and other packages.)

PR #256, opened 2019-11-24
Release Project map navigator v0.94beta

Dafarkias:
[v0.93]

*FEATURE: 'x' sets edit cursor to first visable measure in arrange/track view (a.k.a. Nantho's request)
*Reconfigured marquee. Attempted to bring back its intended behavior/functionality
*Stereo display for media items (previously, only mono. Once again, I used code from spk77 to make this happen.)
*Reconfigured MIDI-items notes visually to be slightly more prominent.
*Removed small bugs, and bolstered script efficiency.

[v0.94] (hot-fix)

*Worked on misrepresentation of stereo items in display

### DIFF
--- a/Various/Dfk_Project Map Navigator.lua
+++ b/Various/Dfk_Project Map Navigator.lua
@@ -1,27 +1,36 @@
 -- @description Project map navigator
 -- @author Dafarkias
--- @version 0.92
--- @changelog Added borderless window option back in (shift+Lclick to center window on mouse)
+-- @version 0.94beta
+-- @changelog
+--   [v0.93]
+--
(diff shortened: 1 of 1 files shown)


cfillion:
The `@changelog` tag should only contain the changes for the current version. You can upload 0.93 and 0.94beta separately to make both available to users.

Because 0.94beta is a pre-release, it is not installed by default unless users opt-in. This means most users will remain at 0.92 and not have access to 0.93 (because as far as ReaPack knows, 0.93 doesn't exist).

Uploading 0.93, then 0.94beta in another PR (or commit) will fix this.

Dafarkias:
Ummmm. Had no clue!

I've actually been considering this script to be in 'alpha,' so I suppose I must not have read the uploading/posting rules very thoroughly. Assuming there are any. 

My apologies!

PR #280, opened 2020-01-24
Release Create a single region with name and tail length from selected items v1.0

EddiePacheco:


cfillion:
Script releases usually prepend a short "author prefix" to the filenames to easily identify the developer in the Action List. It's often a username, shortened version of a name, initials or some abbreviation.

What would you like to use, or do you prefer not using an author prefix for your release(s)?

Additionally it might be good to mention spk77's original script by name in the documentation, which is "Create regions (with tail) from selected items".

EddiePacheco:
> Script releases usually prepend a short "author prefix" to the filenames to easily identify the developer in the Action List. It's often a username, shortened version of a name, initials or some abbreviation.
> 
> What would you like to use, or do you prefer not using an author prefix for your release(s)?
> 
> Additionally it might be good to mention spk77's original script by name in the documentation, which is "Create regions (with tail) from selected items".

Sounds good! Thanks for checking. Let's add "Create regions (with tail) from selected items" in the doc after mentioning spk77. "epacheco" should work as an author prefix, if that's possible.

EddiePacheco:
> > Script releases usually prepend a short "author prefix" to the filenames to easily identify the developer in the Action List. It's often a username, shortened version of a name, initials or some abbreviation.
> > What would you like to use, or do you prefer not using an author prefix for your release(s)?
> > Additionally it might be good to mention spk77's original script by name in the documentation, which is "Create regions (with tail) from selected items".
> 
> Sounds good! Thanks for checking. Let's add "Create regions (with tail) from selected items" in the doc after mentioning spk77. "epacheco" should work as an author prefix, if that's possible.

Ok, the two changes were made now.

PR #283, opened 2020-01-26
Update spk77_Adjust volume envelope point at mouse cursor via mousewh…

amagalma:
added Tempo envelope support


cfillion:
@amagalma The `ReaScript Name` tag must not contain the author name prefix nor `.lua`. Also the changelog is broken in this PR. The proper tag should be used instead of the special (and fragile) wordpress compatibility syntax.

I fixed the metadata (including formatting of the documentation) of your recent three releases in 795eb39bbbd07543f1d716ba6f3ff0d521e4883d. In the future a repository-wide metadata cleanup, that was initiated in #78, is likely to happen.

I highly recommend using https://reapack.com/upload/reascript instead of manually writing the metadata header and submitting PRs by hand.

PR #285, opened 2020-01-29
Release random track order for selected tracks by Will Develop v1.0

Will-Develop:


cfillion:
I fixed the following packaging issues:

- Removed the author name from the package name (no need to write it twice, the Author field is already searchable by users)
- Capitalized the package name (sentence case)
- Removed the version number from the filename (so that future version will have the same command ID within REAPER)
- Fixed a typo in the file name: was "Will-Devolop" instead of "Will-Develop"

Will merge tomorrow if you're OK with these edits.

PR #298, opened 2020-02-15
Release jh_Drop named marker v1.0

johnharkinz:


X-Raym:
@johnharkinz Hi thx for your PR !

I wonder if this pack isn't too specific for this repo, as the needs of a particular named marker is... particular. Some may prefer regions, some prefer CamelCase, some other language, some insert at edit cursor position... A bit unsual that it works on play pos too.

I think in this case a better script would be a script generator which lets generate action with Positio' and Marker type and Name. I wonder if this already exist, but it is possible to make.
Meanwhile I think this PR should be closed without merging. What do you think @cfillion ?

Thx for tour contrib anyway :)

Markers/jh_Drop named markers (bundle).lua
```diff
@@ -0,0 +1,26 @@
...
+--
+--   For example make a copy of a script, update this code in the script file with the name you want to give your marker, rename the script so it makes sense to you, load it in to the actions list, give it a keystroke and then just just play your project and drop meaningful markers.
+--
+--   --
+--   -- CUSTOM MARKER NAME HERE
+--   --
+--   local name = "Bridge"
+--   --
+
+-- @noindex
```
cfillion, line 26:
```suggestion
-- @metapackage
```

Markers/jh_Drop named markers (bundle).lua
```diff
@@ -0,0 +1,26 @@
+-- @description jh_Drop named marker
```
cfillion, line 1:
```suggestion
-- @description Drop named marker
```

cfillion (commented):
At minimum the changes below are needed (no author prefix in ReaPack description, metapackage to not install a file without code).

> I think in this case a better script would be a script generator which lets generate action with Positio' and Marker type and Name. I wonder if this already exist, but it is possible to make.
> Meanwhile I think this PR should be closed without merging. What do you think @cfillion ?

I agree, the preset marker names are too specific without easy customization options.

X-Raym:
@cfillion I propose we close this FR then, and hope for a more flexible tool to come out.

PR #390, opened 2020-05-04
Release JR Snap pitchbend envelope points to semitones negative 24 steps v1.0

J-Reverb:


cfillion:
Considering these four scripts are likely intended to be used together, perhaps it would make more sense to package them as a single package rather than individually?

(Side discussion unrelated to releasing this: Would it make sense to merge the AI variant with the non-AI variant? Unless there is a use case for snapping only the AIs. Some users might prefer scripts that only affect selected points instead of their entire envelope.)

(Code-wise, again unrelated to releasing this, the very long if-else block could be simplified into a short loop to improve the quality of the script.)

J-Reverb:
Thanks for the info, yes I had wondered why I ended up with seperate packages... my bad.
I would appreciate it if you could check the new one and give me the thumbs up that everything is as expected.
Code wise, I will have a go at merging the two scripts as agreed this would be preferable, however I needed to keep the if/else block for the snapping to work correctly as I tried incrementing but ran into a few problems with point values not behaving as expected. Any pointers would be much appreciated on that front. Many thanks.
Also I am re naming the scripts to include tcpfx envelope as they are not intended for volume etc. to save confusion.

PR #426, opened 2020-05-27
Release MK Slicer v1.4.6

alekseymaz:
+ Added an error message if the item or time selection is too short.
+ Improved slicer accuracy. Now the location of the cuts corresponds to what is displayed in the script.
+ Bugfix: now the script ignores the Time Selection if the option is disabled in the settings.


cfillion:
The code changes are missing in this pull request.

PR #434, opened 2020-06-14
Release epacheco create single region with tail length from selected items v1.1

EddiePacheco:
Bug fix for region not accounting for first item not having lowest timeline value. Region created will now properly address item positions in selection.


EddiePacheco (commented):
Changes added are lines 78 through 81 in the updated code.

PR #456, opened 2020-07-27
Distribute MIDI notes evenly : First version

BenTalagan:
A serie of actions for distributing selected notes evenly in time. This is a bit different from stretch/quantize/etc. Initiating forum thread (in french) : https://forum.cockos.com/showthread.php?t=240332 .


cfillion:
Considering the variants grid size, last param and param are trivial, this could be released as a single file/multiple action package.

The main script/package file can use `reaper.get_action_context` to read the action's filename and do the desired operation.

```
@metapackage
@provides
  [main=main,midi_editor] . > talagan_Distribute MIDI notes evenly (grid size).lua
  [main=main,midi_editor] . > talagan_Distribute MIDI notes evenly (param).lua
  [main=main,midi_editor] . > talagan_Distribute MIDI notes evenly (last param).lua
```

BenTalagan:
Ah ok! I was not aware at all of that feature. Very nice! I used a different design pattern, having in mind to avoid duplicating some code on the target installation (hence the lib file). If I understand correctly, reapack duplicates the single file of the package, each copy behaving differently depending on its name. I see pros and cons in the two approaches ; just for info, is there a decisive reason to use the metapackage method instead of the other one? Maybe to prevent the repo from growing too much?

cfillion:
Both methods are valid, so it ends up mostly being a matter of developer preference. (For my own scripts, when the method above makes sense, I like only having to edit and upload one single file.)

Having the actions self-contained when practical makes it easier to tweak the shared code from the Action List's Edit button, or copy to another install w/o ReaPack... At the cost of duplication on the user's end.

(While still retaining all the advantages of a single package)

BenTalagan:
Ok, that really makes sense! It gives me another idea which is an hybrid one that I have to try, I'll propose you a modification of this PR tomorrow. Thanks a lot for the explanations!

BenTalagan:
Ok, I have changed the way the library and the actions interact using a hybrid design pattern : the package is now a metapackage but still installs the lib file separately. The generic action file as a minimal generic code that directly calls the library for handling the parameters passed in the action file name. I have thus added more flavours of the action, and it can be extended by just copy/pasting the action file and renaming it. So in that configuration : 

- Only 2 files will ever exist in the repo (the generic action and the lib file)
- Almost no code redundancy exists on the target installation
- Custom derived actions can be easily added by the user (by duplicating & renaming any action file)

The only drawback is that action files are not totally independant (you have to copy the lib file alongside if you're not using reapack), and the "edit" button does not lead you directly to the core source code of the lib.

BenTalagan:
Ah, another pro for this pattern : if a user duplicates an action to use a custom parameter, he/she will benefit from the library update in case of bug fix/enhancement :). Anyway, on my side I think the PR's good to go now.

PR #459, opened 2020-07-30
Release Remove muted track sends from selected tracks v1.0

Edgemeal:
I deleted the "selected track" script, only posting for others, I don't need it. ;)


cfillion:
This one could be packaged together with "Remove muted track sends" since it's very similar. I don't think it needs to be a separate package.

It can either be added as another file as-is, or the "main" file can set to be installed twice (then the code detect its install filename to select between all vs selected tracks).

Edgemeal:
Please just upload as is.

cfillion:
OK, I've released the non-"selected tracks" one as-is, but what about this one? Your previous commit above has deleted the script (leading me to believe you wanted to re-work it before release).

Edgemeal:
It should be OK, Do I need to re-upload it thru reapack.com?

PR #469, opened 2020-08-11
Update amagalma_Move edit cursor to next possible bad or missing item…

amagalma:
…s crossfade on selected tracks.lua


cfillion:
@amagalma When scripts provide identical functionnality, such as previous/next actions, both actions should be packaged as a single package. This would ensure all users have both installed and of the same version.

PR #514, opened 2020-10-06
Release Save selected tracks floating fx windows (4 slots) v1.00

Edgemeal:


Edgemeal:
Trying to upload script with slots, 1 to 4, did I do it correctly?

cfillion:
The metapackage and provides tags are correct, but they are below the metadata header rather than inside it (looks like you added them by hand?). I fixed it and prepended "edgemeal_" to the install filenames.

Here's how to do it with the GUI: licecap

Edgemeal:
Thank you! Ya I did it by hand, I'll try using the tool next time.

PR #516, opened 2020-10-07
fixed bug in pandabot_Extend items by note divisions, wasn't respecting playrate

benjohnson2001:


cfillion:
The version number must be increased for existing users to receive the update (+ changelog).

PR #531, opened 2020-10-22
Release Turn items into midi v1.1

inmare:
I made script which turn items into midi.
I tried to upload this on here, but I found little error on script.
So I delete that and upload this again with added feature.


cfillion (commented):
The script uses the selected track in the arrange view, not the item currently opened in the MIDI editor. Is there a use case for having it in the MIDI Editor section of the Action List (in addition to the Main section)?

Furthermore, I would suggest adding that to the name to make the script easier to find with the "pitch" and "selected track" keywords. Perhaps something along the lines of "Turn item pitch in the selected track into MIDI notes"?

inmare:
Oh, I think that's more proper name for script. Thanks for advice. I'm new to Reaper and ReaPack both. Then should I just fix the name of the script on here, or upload the whole script again with more proper name and section?

inmare:
I changed the categories to "Item Properties", and changed the file name to "yatsumehole_Turn item pitch in the selected track into MIDI notes.lua" . Is it okay to change?

PR #532, opened 2020-10-24
Release Cycle item last selected take up-down v1.00

Edgemeal:


cfillion:
I made a small edit in 4481df5f06a183285608a4df4dad2afee20011aa to only install the two up and down actions. To do this with the GUI: set the package file's resource type to "Don't install".

Edgemeal (commented):
I thought I used upload tool correctly this time, I guess not. DAMN IT!

Edgemeal:
> I made a small edit in [4481df5](https://github.com/ReaTeam/ReaScripts/commit/4481df5f06a183285608a4df4dad2afee20011aa) to only install the two up and down actions. To do this with the GUI: set the package file's resource type to "Don't install".

THANKS!

PR #535, opened 2020-10-25
updated dock state persistence method for ChordGun

benjohnson2001:
atexit didn't work as I expected https://forum.cockos.com/showthread.php?t=243978


benjohnson2001:
ah hold on, there's a bug. I'll come back in a bit with version 1.6

benjohnson2001:
okay should be all set now

PR #536, opened 2020-10-27
Release Save ripple mode to persistent ExtState v1.0

bowlneudel:


cfillion:
Both save and restore actions should be bundled in the same package because they are intended to be used in conjunction.

bowlneudel:
> Both save and restore actions should be bundled in the same package because they are intended to be used in conjunction.

Agreed cfillion!

I realized after I submitted them separately that they would crowd the package browser unnecessarily. However, I'm completely new to github, reascript, and coding in general.

I'll investigate how to combine multiple reascripts into a package – is a new pull request required, or can I simply reorganize the existing branches?

If you have the capacity and the permissions, you are more than welcome to package them on my behalf.

X-Raym:
@cfillion I think we can merge theme separately anyway,
though I wonder if the naming isnt too specific.

Maybe just Save ripple state.lua will be better than explicitely talking about extstate, which isnt a common term in action list.

cfillion:
Both actions are required for the script's basic functionality (installing just the 'save' action without 'restore' or vice-versa is useless), so they really should be packaged together.

X-Raym:
@cfillion the author definitely need assistance for this then cause he doesnt seems to have succeed on this side

bowlneudel:
Hey it's me, the author. I haven't become any more proficient at packaging scripts in the interim. Any assistance here would be most welcome, and likewise renaming these actions to something more conventional.

PR #573, opened 2020-12-10
Release unsolo track with its number v1.0

menaAwad:


Tracks Properties/menaawad_Unsolo track by its number.lua
```diff
@@ -0,0 +1,26 @@
...
+ sel_tr = reaper.GetSelectedTrack(0, 0)
+if  sel_tr ~= nil then
+ sel_tr_num = reaper.GetMediaTrackInfo_Value(sel_tr, "IP_TRACKNUMBER")
+retval, preset = reaper.GetUserInputs("EnterTrackNumberToUnsolo", 1, "", sel_tr_num)
+else
+retval, preset = reaper.GetUserInputs("EnterTrackNumberToUnsolo", 1, "", "0")
+end
+
+if (preset == "0" or preset == "") then return end
+if (tonumber(preset)  >= 0 and tonumber(preset) <= num_of_tracks) then
```
cfillion, line 17: There are a few bugs:

- The script crashes here if the user input isn't convertible to string (eg. 'foobar')
  
      attempt to compare number with nil
- It crashes at the `reaper.GetTrack(0, t)` call below if the user inputs a negative number or a decimal number

      bad argument #2 to 'GetTrack' (number expected, got nil)
      bad argument #2 to 'GetTrack' (number has no integer representation)
- It unsoloes the selected track even if the user chooses Cancel in the prompt dialog
- The selected track number is shown as a decimal number in the prompt
- The prompt does not contain text

Possible fix:

```lua
local num_of_tracks = reaper.CountTracks(nil)
if num_of_tracks == 0 then return end

local sel_tr = reaper.GetSelectedTrack(nil, 0)
local sel_tr_num = sel_tr and reaper.GetMediaTrackInfo_Value(sel_tr, 'IP_TRACKNUMBER') or 0
sel_tr_num =  math.tointeger(sel_tr_num)

local script_name = ({reaper.get_action_context()})[2]:match('([^/\\_]+)%.lua$')
local retval, user_input = reaper.GetUserInputs(script_name, 1, 'Track number:', sel_tr_num)
local track_number = math.tointeger(user_input)
if not retval or not track_number or track_number < 1 or track_number > num_of_tracks then
  return
end

local tr = reaper.CSurf_TrackFromID(track_number, false)
reaper.SetMediaTrackInfo_Value(tr, 'I_SOLO', 0)
reaper.Main_OnCommand(reaper.NamedCommandLookup('_OSARA_REPORTSEL'), 0)
```

menaAwad:
I have two questions please how can I play sound file from LUA and other question is that how can I get shortcut of action by eights it’s action ID

PR #577, opened 2020-12-16
Fix Lokasenna_Create folder to contain selected tracks.lua

leafac:
[The current implementation includes the following line](https://github.com/ReaTeam/ReaScripts/blob/7c05f6416cc0625117250cda6264d77053daf321/Tracks%20Properties/Lokasenna_Create%20folder%20to%20contain%20selected%20tracks.lua#L44):

```lua
local newDepth = (curDepth == 0) and (curDepth - 1) or -1
```

This line doesn’t make sense, because if `curDepth == 0` holds, then `curDepth - 1` is `-1`, so `newDepth` is always `-1`.

I believe the intent is to say `newDepth = curDepth - 1`, which means the that the `lastSel` is now the last track of yet another folder (the one we just created).

With this fix, the action behaves more sanely when creating folders within folders. Consider the following example:

```
TRACK 1
  TRACK 2 (SELECTED)
  TRACK 3 (SELECTED)
TRACK 4
```

In the current implementation the result would be:

```
TRACK 1
  TRACK NEW
    TRACK 2 (SELECTED)
    TRACK 3 (SELECTED)
  TRACK 4
```

But with my fix the result is:

```
TRACK 1
  TRACK NEW
    TRACK 2 (SELECTED)
    TRACK 3 (SELECTED)
TRACK 4
```

Note how `TRACK 4` hasn’t been nested into `TRACK 1`.

There’s still (both in the original version as well as in my fixed version) an inconvenience when the last selected track is itself a folder, for example:

```
TRACK 1 (SELECTED)
TRACK 2 (SELECTED)
  TRACK 3
```

Which results in:

```
TRACK NEW
  TRACK 1 (SELECTED)
  TRACK 2 (SELECTED)
TRACK 3
```

Note how `TRACK 2` is no longer a parent of `TRACK 3`.

But fixing this would require traversing `TRACK 3` onward looking for the last child. This would complicate the action beyond its scope (I think, if you disagree let me know and I can try to come up with a pull request). One can always select `TRACK 3` as well and then it works.

### DIFF
--- a/Tracks Properties/Lokasenna_Create folder to contain selected tracks.lua
+++ b/Tracks Properties/Lokasenna_Create folder to contain selected tracks.lua
@@ -41,7 +41,7 @@ reaper.SetMediaTrackInfo_Value( parent, "I_FOLDERDEPTH", 1, true )
 local lastSel = reaper.GetSelectedTrack(0, reaper.CountSelectedTracks(0) - 1)
 
 local curDepth = reaper.GetMediaTrackInfo_Value( lastSel, "I_FOLDERDEPTH" )
-local newDepth = (curDepth == 0) and (curDepth - 1) or -1
+local newDepth = curDepth - 1
 
 reaper.SetMediaTrackInfo_Value( lastSel, "I_FOLDERDEPTH", newDepth, true )
 
--- a/Tracks Properties/Lokasenna_Create folder to contain selected tracks.lua
+++ b/Tracks Properties/Lokasenna_Create folder to contain selected tracks.lua
@@ -1,10 +1,13 @@
 --[[
 Description: Create folder to contain selected tracks
-Version: 1.0.0
+Version: 1.1.0
 Author: Lokasenna
 Donation: https://paypal.me/Lokasenna
 Changelog:
-	Initial release
+* v1.0.0
+   + Initial release
+* v1.1.0 (2020-12-23)
+  + Fix folder creating when last selected track is last in folder or a folder itself.
 Links:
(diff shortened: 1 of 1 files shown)


cfillion:
@jalovatt Is this change OK to merge for you?

@leafac Assuming Lokasenna agrees with the change: please increase the version number and update the changelog in the header at the top of the file.

jalovatt:
I'm not able to test it right now, but that looks fine to me.

Thanks for catching that, @leafac

leafac:
@cfillion: I updated the documentation as requested.

@jalovatt: Thank you for reviewing my contribution as well as for everything else you did. I learned a lot from reading your scripts and forum posts. It’s what made [this script](https://youtu.be/hSnk6ueU3hQ) possible 😃

**Both:** I bit the bullet and fixed the other case mentioned in my original message above. Now the script does the right thing (that is, it doesn’t destroy the existing folder structure) even if the last selected track is a folder. You may want to take a brief look at this before merging.

leafac:
Thanks for merging my contribution 😃

For people who end up finding this, I want to mention two things:

1. If you don’t like how this action asks for a parent name, or if you want a dual action to ungroup tracks from folders, then check out my [`leafac_Group tracks into a folder.lua` and `leafac_Ungroup tracks from within folders.lua`](https://youtu.be/ap1IwhXh4Uo).

2. If you want to learn more about how my contribution works, check out [the code review for my actions mentioned above (which contain pretty much the same snippet)](https://youtu.be/HDtZPWM9gpY).

PR #587, opened 2020-12-24
Release Convert filtered custom actions to LUA ReaScripts or dump as LUA code v1.0

Buy-One:


cfillion:
I recommend packaging closely related scripts in a single package. There could be a "Convert custom action to LUA ReaScript" package containing all three actions, always installed/updated together. This also allows the scripts to share common code.

PR #592, opened 2020-12-30
Update spk77_Export item names and time stamps to text file.lua - Add Project Time Offset to Position

EricHeidbreder:
Script was returning values based on the beginning of the project. Added the project time offset to position to account for projects with an offset. This change does not affect projects with no offset.


cfillion:
Please increase the version number and replace the changelog.

EricHeidbreder:
Version number and changelog updated!

PR #597, opened 2021-01-03
Release Convert custom actions to Lua ReaScript v1.2

Buy-One:


Buy-One:
@cfillion  Uploaded 3 files as additional files in the metapackage. They're not in the branch. Not sure where to look for them and whether they have been uploaded.  Should they be uploaded directly into my repository?

Various/BuyOne_Convert custom actions to Lua ReaScript.lua
```diff
@@ -0,0 +1,14 @@
+-- @description Convert custom actions to Lua ReaScript
+-- @author BuyOne
+-- @version 1.2
+-- @metapackage
+-- @provides
+--   [main] . > BuyOne_Convert custom action to Lua ReaScript.lua
+--   [main] . > BuyOne_Convert custom actions in batch to Lua ReaScripts or dump as Lua code.lua
+--   [main] . > BuyOne_Convert filtered custom actions to Lua ReaScripts or dump as Lua code.lua
```
cfillion, line 8: This installs the Package file three times under different names. This technique is used when more than one script shares the same code. A script can detect its own filename at runtime using `({reaper.get_action_context()})[2]:match("([^/\\_]+)%.lua$")` and adapt its behavior. My "Move left/right edge of selected automation items to start/end of time selection" package uses this, for example.

However, because there is no code below, the three script files wouldn't do anything once installed.

Buy-One: Got it.

PR #598, opened 2021-01-04
Release Convert custom actions to Lua ReaScript v1.2

Buy-One:


Various/buyone_Convert custom actions to Lua ReaScript/BuyOne_Convert custom actions in batch to Lua ReaScripts or dump as Lua code.lua
```diff
@@ -0,0 +1,391 @@
...
+
+-- Concatenate the path
+	if subdir ~= '' and subdir ~= 'existing or new; or leave as it is or empty' then
+	-- check if such directory already exists and if not, create
+		for dir in io.popen('dir \"'..path..sep..'Scripts\" /ad'):lines() do
+			if dir == subdir then exists = true break end
+		end
+		if not exists then os.execute('mkdir \"'..path..sep..'Scripts'..sep..subdir..'\"')
+		-- check if the directory has been created
+			for dir in io.popen('dir \"'..path..sep..'Scripts\" /b'):lines() do
```
cfillion: This bit is also vulnerable to command injection and won't work on stock macOS (no `dir` is provided). You can use `RecursiveCreateDirectory` from REAPER for creating directories.

Otherwising, this is correctly packaged. Will merge this after the calls to `os.execute` or `io.popen` are fixed.

---

`os.execute` and `io.popen` sends a script to the shell for running commands. The shell is a full scripting language interpreter, so directly injecting outside data (user input, file paths) into a shell script (the "command line") is opening a door for bugs to happen.

Answering your question in a previous PR about "sanitizing": this mean ensuring all characters with special meaning to the current system's shell will be interpreted literally. It's somewhat risky and shell syntax varies from system to system.

Sanitizing for `sh` and derivatives (macOS and most Linuxes's default) can be as simple as using single quotes and escaping single quotes within the argument: `'Lua'\''s : operator is nice'`.

But it's better to avoid invoking the system shell when possible. `reaper.ExecProcess` does the argument splitting itself without going to the shell. It supports ' " and ` for quoting values containing spaces, but has no escape character (meaning values combining space and 2+ types of quotes are not supported).

Buy-One: OK. Could you give an example of the kind of harm which can be caused to the user system via these commands by the user themself? Or do you mean it may happen by accident?

cfillion: > Or do you mean it may happen by accident?

^ This. Realistically, the worse that is likely to happen naturally is the feature wouldn't work (or behave strangely) with some input/filenames.

(Anything *can* happen though: https://i.imgur.com/hkgL0mB.gif.)

Buy-One: OK. Could you give an example of the kind of harm which can be caused to the user system via these commands by the user themself? Or do you mean it may happen by accident due to variations of shell syntax on different systems?

Buy-One:
@cfillion I think it's ready now. Thank you for assistance.

Various/buyone_Convert custom actions to Lua ReaScript/BuyOne_Convert custom actions in batch to Lua ReaScripts or dump as Lua code.lua
```diff
@@ -0,0 +1,390 @@
...
+	elseif flags_t[2] then inset = 'MIDI ' end
+	if flags_t[1] and flags_t[2] and flags_t[5] then inset = 'Main & MIDI from Arrange '
+	elseif flags_t[5] then inset = 'MIDI from Arrange ' end
+
+local pref, f_name, ext = '', inset..tostring(#lines_t)..' custom actions LUA dump', '.txt' -- name pattern for individual LUA files is concatenated inside the MAIN LOOP since it depends on each custom action properties
+
+
+-- Check if there's already a txt dump file in the target directory and ovewrite if confirmed by the user
+	if flags_t[4] then
+		for f in io.popen('dir \"'..f_path..sep..'\" /b'):lines() do
```
cfillion: Another unprotected `io.popen` here. A simple way to check whether a file exists and is readable:

```lua
function file_exists(fn)
   local f = io.open(fn, 'r')
   return f ~= nil and f:close()
end

file_exists(f_path..sep..f)
```

Buy-One: Oops, will fix.

Buy-One: @cfillion Done.
Could you clarify the line `return f ~= nil and f:close()`? Does it return true if `f` isn't nil? And is it a kind of ternary operator where the file is closed only if `f` is NOT nil? And if it's nil, is it not closed or is there simply nothing to close?

cfillion: All of the above. It returns true only if `f` is not nil and it was closed successfully. The `f:close()` is skipped if `f` is nil.

It's the same as:
```lua
local opened_succesfully = f ~= nil

if opened_succesfully then
  return f:close()
else
  return opened_succesfully
end
```

Buy-One: Got it, thank you.

PR #617, opened 2021-02-08
BuyOne_Move pitch cursor

Buy-One:


Buy-One:
Before merging would like to double check one thing. I see this pull request includes files from the previous one already merged. They haven't changed on my side, so they're dated earlier than the date of the last merge (i suppose). Will they be merged as well or will they be automatically ignored as outdated? And if they will, to avoid re-merging them should i delete them from my branch?

Buy-One:
Before merging would like to double check one thing. I see this pull request includes files from the previous one already merged. They haven't changed on my side (as far as i recall), so they're dated earlier than the date of the last merge (i suppose). Will they be merged as well or will they be automatically ignored as outdated? And if they will be merged, to avoid re-merging them should i delete them from my branch?

cfillion:
I'm squashing the commits when merging so there will only be one commit. It would be best to update your fork to match this repository's history to avoid this though. I would advice using a different branch than master when doing changes (there is no way to sync your fork's master branch with upstream's online using GitHub's UI, short of recreating the fork). (Or using reapack.com/upload which takes care of all of that automatically.)

Buy-One:
I'm not quite versed in the Github lingo. Do i understand correctly that for this request it can be left as it is? For future requests, i understand the requirement of updating the fork.
Regarding your advice of  `using a different branch than master when doing changes`, does it mean that files can be merged with the main repository even if they come from an independent repository, not forked from the main? Or if i create a separate  branch within the fork specifically for a new upload, will the pull request only contain the files from that branch.

P.S. I think i got it, new files need to be uploaded via the main repository so a new branch is created from there. If i'm right, only tell me if i should close this pull request and resubmit.

PR #643, opened 2021-02-21
Create amagalma_Crossfade items in Razor Edit area.lua

amagalma:


amagalma:
Will squash and merge when official build with RE is out

PR #650, opened 2021-02-21
Release Moves items rigth, preserving crossfade lengths v2.0

Rodilab:


cfillion:
These left/right action pairs should provided together in the same package.

PR #654, opened 2021-02-27
Release Preset velocity v0.2

Malfunky:
-- CHANGELOG  ( changelog )  Version 0.2
-- 2-25-2021 - Excessive defer calls have been replaced with tail calls. Script is now faster.
-  Eliminated excessive calls to update()
-  Note Preview.jsfx :  velocity 0 "noteoffs" no longer passes as note ons to cause trouble.


cfillion:
The updated kl_Preset Velocity.lua  has been uploaded as new additional file instead of updating the exiting kl_Preset Velocity.lua  package file.

PR #667, opened 2021-03-04
Release Render whole items in new take (preserve take name) v1.0

Rodilab:


cfillion:
This action should be added to the existing "Render whole items in new take" package.

PR #672, opened 2021-03-08
Release yaunick_Bounce selection (like in Studio One) v1.0

Yaunick:


Yaunick:
Need to add author Yaunick_ or yaunick_? Didn't quite understand at first

PR #679, opened 2021-03-17
Release FX list as popup v1.1

Tagirijus:
Changed the name of the author and the script to be more coherent with my other scripts.


cfillion:
The name hasn't been changed. Is "FX list menu" or "FX list as popup" the correct name?

Tagirijus:
Hm, maybe I did a mistake or some autofil _"magic"_ accidentially happend. The correct name for the script should be:

**FX list menu**

Thank you and sorry again for the inconvenience!

PR #693, opened 2021-03-29
Release MK Slicer v2.11

cool9-MK:
+ Restored the ability to edit the position and velocity of markers (it was broken in the previous version, sorry).
+ Now toggling the Grid buttons does not reset the Swing amount.
+ Now the Swing slider does not show -0 when the value is 0.
+ Quick Q is now available when switching Slice / Markers modes and after changing Grid / Swing. Just press Q to get the result right away. After changing the parameters, pressing Q again will also work.
+ Fixed a bug. Now, when you change the parameters of the sliders using the mouse wheel, the changes will be immediately noticeable when the buttons are pressed.

### DIFF
--- a/Items Editing/cool_MK Slicer.lua
+++ b/Items Editing/cool_MK Slicer.lua
@@ -1,15 +1,12 @@
 -- @description MK Slicer
 -- @author cool
--- @version 2.4
+-- @version 2.11
 -- @changelog
---   + Added a Sync button (zoom and waveform position) with Arrange View window.
---   + Added controls for easy access to changing grid parameters.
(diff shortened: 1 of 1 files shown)


cfillion:
ReaPack version numbers are not decimal (they're an infinite list of integer and alphabetic components). Because 2.1 from the previous PR is lower than 1.03, I renamed it to 2.4 for the update to be distributed.

While 2.11 is valid (higher than 2.4), releasing this will prevent any versions from 2.5 to 2.10 from being released as updates in the future. Patch updates should increment a third integer component: eg. 2.4.1.

cool9-MK:
Got it. Thank you!

PR #697, opened 2021-04-01
BuyOne_Insert selected FX in OR copy focused FX to selected ob…

Buy-One:


Buy-One (commented):
I'd like to keep the second header for after the script is downloaded, so the info is readily available to the user. It doesn't cause conflict since it's commented out, correct?

cfillion:
It repeats the same data as the reapack-index header though (except for the min. reaper version and license, I didn't notice those). The reapack-index header is not stripped out when installed.

Buy-One:
It does with some variations and is meant for the user, because once the script is installed it's not viewed via ReaPack interface so the ReaPack info isn't readily available. It doesn't harm, does it?

Buy-One:
Yep, i'm slow. My brain got stuck on metapackages which i've been submitting lately.

Buy-One:
Yep, i'm slow. My brain got stuck on metapackages which i've been submitting lately. So as far as i'm concerned it's good to be merged.

PR #702, opened 2021-04-09
Update spk77_Sexan_Folder Record and Monitor Arm To Children.lua

Buy-One:
Updated according to Sexan's instructions from Sep 2020
https://forum.cockos.com/showpost.php?p=2344559&postcount=18


Buy-One:
Wonder if it will be merged. Or does it depend on the authors?

PR #708, opened 2021-04-18
Release Color palette v1.3

Rodilab:
Update : Dock mode, Vertical mode, Any modifier-click on "Item" button switch to "Track Name" mode and viceversa, Ctrl-click set color in children tracks (Ctrl+Alt in Windows), Background color like Default_6.0 theme, Save last window position, Bug fixed when drag and drop and right click.

### DIFF
--- a/Various/rodilab_Color palette.lua
+++ b/Various/rodilab_Color palette.lua
@@ -1,13 +1,17 @@
 -- @description Color palette
 -- @author Rodilab
--- @version 1.2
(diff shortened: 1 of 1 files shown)


cfillion:
Just a heads up: the next version of ReaImGui (0.3) will add a new API for (un)docking contexts, making the current way of using the REAPER dock functions unsupported (calling DockWindowAdd* mid-frame is crash-prone on Linux).

PR #719, opened 2021-04-23
Release Color palette v1.51

Rodilab:
- Track Manager and Region/Marker Manager work on all OS
- Ctrl+Alt-click (Ctrl-click on Mac) with 'Takes' target set color to all takes (active, and non actives)
- Highlight border is black if color is very light
- Improved "Set 1st user to default tracks" option
- Improved "Track Manager and Region / Marker Manager" detection

### DIFF
--- a/Various/rodilab_Color palette.lua
+++ b/Various/rodilab_Color palette.lua
@@ -1,13 +1,12 @@
 -- @description Color palette
 -- @author Rodilab
--- @version 1.5
+-- @version 1.51
 -- @changelog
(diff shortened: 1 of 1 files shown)


cfillion:
ReaPack version number are not decimal. Releasing a v1.51 will mask any future v1.6 - v1.50 updates. v1.5.1 might be a better choice in this case...

Rodilab:
Okay, thanks ;)
Do I have to re post it? Or are you editing this one ?

PR #723, opened 2021-04-27
Update BuyOne_FX presets navigation.lua

Buy-One:


cfillion:
Looks like there is no changes to merge? (Beside extra newlines, which would break the header as parsing stops at the first empty line.)

Buy-One:
My aim was to place each script name on its own line.

I've added dashes on the empty lines, hopefully this is the correct formatting.

PR #737, opened 2021-05-11
Update and rename Load FX chain on selected objects (guide inside)

Buy-One:
Not sure if i should bump the version.


cfillion:
> Not sure if i should bump the version.

Renaming is the same as deleting the old package and uploading a new one, so bumping the version is not necessary.

PR #744, opened 2021-05-19
BuyOne_Dynamic ReaperMenu (guide inside).lua

Buy-One:


Various/BuyOne_Dynamic ReaperMenu (guide inside).lua
```diff
@@ -0,0 +1,254 @@
...
+Changelog: Initial release
+Provides: [main=main,midi_editor] .
+About:
+
+#### GUIDE
+
+- The script is designed to allow using REAPER menus independently of REAPER
+menu system by loading .ReaperMenu files on demand, which essentially allows
+potentially having and using many more menus and toolbars than the REAPER
+native menu system provides.
```
cfillion: The content of this `About` tag is empty, because the following text is not indented.

Buy-One: Thank you, by how much should it be indented?

cfillion: The amount of spaces or tabs doesn't matter, as long as it is consistent across all lines. I'd recommend using the package editor instead of manually writing the metadata and pull requests...

Buy-One: OK, i think i'm more comfortable with this interface.

PR #747, opened 2021-05-19
Release Color palette v1.8.1

Rodilab:
Updated due to the new version of ReaImGui v0.4.0


cfillion:
```lua
FLT_MIN = 1.17549e-38
```

That approximation doesn't work properly on Windows. v0.4 adds `NumericLimits_Float()` for getting the exact values of FLT_MIN and FLT_MAX.

PR #752, opened 2021-06-02
Release MK Shaper/Stutter v1.0

cool9-MK:
Initial Release

### DIFF
--- a/Envelopes/cool_MK ShaperStutter.lua
+++ b/Envelopes/cool_MK ShaperStutter.lua
@@ -1,0 +1,7140 @@
+-- @description MK Shaper/Stutter
+-- @author cool
+-- @version 1.0
+-- @changelog Initial Release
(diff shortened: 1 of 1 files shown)


Envelopes/cool_MK ShaperStutter.lua
```diff
@@ -0,0 +1,7140 @@
...
+local Grid_Fill_TB = {elm_table[18]}
+
+---------------------------------------------------------------
+---  Create Menu Settings   -----------------------------
+---------------------------------------------------------------
+
+function OpenURL(url)
+  local OS = r.GetOS()
+  if OS == "OSX32" or OS == "OSX64" then
+    os.execute('open "" "' .. url .. '"')
```
cfillion, line 3098: This opens an extra Finder window (due to the extra first argument to `open`) in addition to opening the URL. (And this `OpenURL` doesn't support Linux).

I'd recommend using `CF_ShellExecute` from SWS instead, which is cross-platform.

PR #770, opened 2021-06-27
Release Color palette v2.0

Rodilab:
- New "Arm" mode
- New ReaImGui v0.5 window
- Auto-resize cells
- Arial font
- Background color matches Reaper background color (can be disabled)
- New widget for non-linear brightness/saturation
- New button "Settings" with icon


Rodilab:
This is for ReaImGui v0.5.0 update

PR #773, opened 2021-07-01
Release Color palette v2.10

Rodilab:
- Minor code improvement
- Fix popup window position

### DIFF
--- a/Various/rodilab_Color palette.lua
+++ b/Various/rodilab_Color palette.lua
@@ -1,14 +1,9 @@
 -- @description Color palette
 -- @author Rodilab
--- @version 2.0
+-- @version 2.10
 -- @changelog
---   - New "Arm" mode
---   - New ReaImGui v0.5 window
---   - Auto-resize cells
---   - Arial font
---   - Background color matches Reaper background color (can be disabled)
---   - New widget for non-linear brightness/saturation
---   - New button "Settings" with icon
+--   - Minor code improvement
+--   - Fix popup window position
 -- @provides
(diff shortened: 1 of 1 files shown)


Various/rodilab_Color palette.lua
```diff
@@ -1,14 +1,9 @@
 -- @description Color palette
 -- @author Rodilab
--- @version 2.0
+-- @version 2.10
```
cfillion, line 3: 10 would prevent releasing v2 updates with minor version numbers 1 through 9. Is this OK?

Various/rodilab_Color palette.lua
```diff
@@ -548,6 +543,10 @@ function set_window()
   ctx = r.ImGui_CreateContext(script_name, r.ImGui_ConfigFlags_DockingEnable())
   font = r.ImGui_CreateFont('Arial',12)
   r.ImGui_AttachFont(ctx,font)
+  if conf.mouse_pos then
+    cur_x, cur_y = r.GetMousePosition()
+    cur_x, cur_y = r.ImGui_PointConvertNative(ctx, cur_x, cur_y, true)
```
cfillion: `to_native` must be false (or omitted, false is the default) when converting from platform-native coordinates to ImGui coordinates. (Otherwise HiDPI scaling will be applied in reverse.)

Rodilab:
Hi Christian,

Yes, it is clearer to me.
Using '2.10' '2.20' '2.30' allows me to update a '2.31' for example (instead of '2.3.1').

PR #805, opened 2021-08-13
Release Restore tracks floating fx windows (4 slots) v1.01

Edgemeal:
When restoring a slot and the floating fx window is already open then it is closed to act like a toggle.


cfillion:
The package file got marked for installation instead of being "Do not install"/unticked. I've re-marked it as a metapackage in 62c2623b7635088c48f31c09d409e4c2794c31b8.

Edgemeal:
Thanks, and sorry bout that, I seem to always screw up something, somehow! :/

PR #819, opened 2021-08-28
API help: Undo_EndBlock(): add link to extra flags description (post from schwa)

nofishonfriday:


nofishonfriday:
- It seems to 'assign' MPL as author for my change/commit (https://github.com/ReaTeam/ReaScripts/pull/819/files#diff-413687310c19d4e64433aaaffecfdd0e9b298381e67f93de29087380a95e0bbfR1978), is there a way to supply a dedicated author for changes?
- I'm not sure which versioning scheme we should use here (n.nn, as I've done currently, or n.n.n or ...?)

Buy-One:
@nofishonfriday 

Tried the script, but it doesn't allow scrolling through the entries or switching to those which are outside the window limits. Is this your experience as well?

BTW appreciate the link to schwa's post

nofishonfriday:
@Buy-One 
Admittedly I haven't actually tried the script before submitting this PR. :P (will do later on..)
But if it doesn't work properly for you I think probably best would be to post in the thread from where it evolved (https://forum.cockos.com/showthread.php?t=256796) to let mpl know about it or open a dedicated issue about it here.

cfillion:
Pinging @MichaelPilyavskiy.

> * It seems to 'assign' MPL as author for my change/commit (https://github.com/ReaTeam/ReaScripts/pull/819/files#diff-413687310c19d4e64433aaaffecfdd0e9b298381e67f93de29087380a95e0bbfR1978), is there a way to supply a dedicated author for changes?

Yes, but it changes the displayed ownership of the whole package. Probably shouldn't be done for contributed patches.

In the [package editor](https://reapack.com/upload) (it edits the `@author` metadata tag):

Author edit field

Maybe adding a note next to the changelog line would be better?

nofishonfriday:
@cfillion 
> Yes, but it changes the displayed ownership of the whole package. 

Ok, that's probably not what's desired. :)

> Maybe adding a note next to the changelog line would be better?

Thought about this too, but then again, thinking about SWS, we don't put in the changelog either from whom a change comes.
After all, author of a change doesn't bring much additional value for end user, no?
Anyway, no biggie, I just found it surprising that MPL's name showed up there without being directly involved.
Given the above I'd tend to leave it as is now (not changing displayed ownership and no note about author of change) .

MichaelPilyavskiy:
It is totally ok to list all contributors in the script authors line. I willll remove later links to me and let only forum thread link.
Scroll - well,I will take a look.

nofishonfriday:
@MichaelPilyavskiy
Ok, put my name in authors line.

A couple of questions/thoughts:
- Can we put linebreaks in the DESC lines or need it be all in a single line?
Without linebreaks I guess putting code snippets there might get a bit cluttery?

- I now put 'ReaTeam:' before my annotation (https://github.com/ReaTeam/ReaScripts/pull/819/files#diff-939f80e8a65cbf8864abc17a1bcd23aebe68e8cccb5bdc0669df3d65c69b8066R2256) because I think it's a good idea to make clear what's official doc text and what is additions from us. What do you think?

- Are you ok with the versioning scheme I started (1.01, 1.02 etc.) or do ypu propose something else?

- The parsed ReaImGUI functions text may look a bit odd (the Â in front: https://github.com/ReaTeam/ReaScripts/commit/9dc8f998f41a3198e2306b0a70aae023948bfd14#diff-939f80e8a65cbf8864abc17a1bcd23aebe68e8cccb5bdc0669df3d65c69b8066R2839) or is this correct?

cfillion:
> The parsed ReaImGUI functions text may look a bit odd (the Â in front: 9dc8f99#diff-939f80e8a65cbf8864abc17a1bcd23aebe68e8cccb5bdc0669df3d65c69b8066R2839) or is this correct?

These characters are supposed to be non-breaking spaces in UTF-8.

MichaelPilyavskiy:
https://github.com/ReaTeam/ReaScripts/pull/820

edit: oh, I firstly added ReaTeam as script author, then added nofish, but forgot to edit changelog

nofishonfriday:
Superseded by #820.
Thanks mpl.

PR #822, opened 2021-08-31
Release Disable track playback offset for selected tracks v1.00

Edgemeal:


cfillion:
> Note: This script is not needed, REAPER now has an action to do this!

Maybe it should be deleted from the repository then? Doing so would only prevent new (re)installs without affecting existing users (beyond the obsolete prompt on synchronizing).

Edgemeal:
OK, I'll delete it ,. Not sure how, I view file in guthub from link above, but delete file option is grayed out.

cfillion:
I deleted it via 12498e722aace1757e9b5e3ea988d0b42731958d.

>  Not sure how, I view file in guthub from link above, but delete file option is grayed out.

Either with git locally, or via GitHub by browsing to the file (https://github.com/ReaTeam/ReaScripts/tree/master/Tracks%20Properties). The edit/delete button on this PR are disabled because the PR is merged (not sure if deleting a file from a PR marks it for actual deletion or if it just undoes the changes?).

Edgemeal:
I just refreshed ReaPack and that script disappeared so looks like its deleted now, THANKS!

PR #824, opened 2021-09-01
Release Select next region in region manager window v1.00

Edgemeal:


cfillion:
Next/Previous actions should be packaged together.

Edgemeal:
OK I try again, hope I did it correctly this time

But I think the main title was supposed to just be,
**Select region in region manager window (Next/Previous)**  

Not sure how I can change that now.

PR #844, opened 2021-09-22
Release JRT_Navigate-Zoom Out Vertically OR Move Notes Down Ignoring Key v1.0

JRTaylorMusic:


cfillion:
Previous/Next, Up/Down, In/Out actions should be bundled in pairs rather than as one separate packages for each. This is so they are installed and updated together.

Here's an example using the GUI package editor: https://i.imgur.com/QAgln6k.gif.

PR #849, opened 2021-09-23
Added two new region scripts

Landeplage:
I use these scripts daily. Great for game audio SFX exporting.

### DIFF
--- a/Regions/Mordi_Create single region from selected items (get name and color from folder track) and mark it in region render matrix via master.lua
+++ b/Regions/Mordi_Create single region from selected items (get name and color from folder track) and mark it in region render matrix via master.lua
@@ -1,13 +1,15 @@
--- @description Create single region from selected items (get name and color from folder track) and mark it in region render matrix via master
--- @author Mordi
--- @version 1.0
--- @changelog
--- @provides [main] .
--- @screenshot Creating regions from items https://i.imgur.com/xUg7bSU.gif
--- @about
---   # Create single region from selected items (get name and color from folder track) and mark it in region render matrix via master
---
---   Made for exporting sound effects. Select all the items that make up your sound effect, and run this script. A region will be created which encompasses all the items and inherits its name and color from the parent track of the topmost item.
+--[[
+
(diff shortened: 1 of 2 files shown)


Buy-One:
@Landeplage 

Hey

You seem to have forgotten to add `Provides: [main] .` tag.

Landeplage:
@Buy-One Trying to find documentation on how to use that tag, but not finding any. The two scripts are independent. Do they still need the `Provides` tag?

Edit: Found documentation, but still a bit unclear. Should each PR only consist of a single file if they are independent?
https://github.com/cfillion/reapack-index/wiki/Packaging-Documentation#provides

Landeplage:
@Buy-One Nevermind - was able to figure it out. Thanks for the tip about the Provides tag.

cfillion:
A provides tag is not necessary. The issue was caused by the empty `changelog` tag. Will merge this later.

I recommend using the form at https://reapack.com/upload to automatically generate the metadata and submit the pull requests.

Buy-One:
> A provides tag is not necessary

Does this mean that without explicit Provides tag the script is always installed in the Main section and only if other sections are involved it's necessary?

cfillion:
Yes. The default is to register the script in the Action List's main section (or in the MIDI Editor if it's the folder with the same name).

cfillion:
Looked more closely into it before merging, these two scripts appear to be variations on your existing "Create single region from selected items (get name and color from folder track)" script.

Additional actions should be added to the existing package instead of submitted separately. This can be done using the GUI at https://reapack.com/upload:

1. Type "Create single region from selected items" in the name field and click on your package to load it in the editor
2. Bump the version number of the package
3. Drag/drop both new script files over the "Add file" button
4. For each of those two files, click on the "Action List" drop-down menu and enable registering them as actions in the Main section

Landeplage:
@cfillion Done.

Is there any meaningful distinction between the "Package file" and the "Hosted files"?
https://i.imgur.com/8cookfl.png

cfillion:
The only difference is the "package file" contains the metadata information, which the other files don't have (except for a "noindex" tag).

PR #871, opened 2021-10-18
Release Move selected items content to previous cue (align with snap offset) v1.0

Landeplage:


cfillion:
Same as #849, previous/next actions should be packaged together. https://github.com/ReaTeam/ReaScripts/pull/849#issuecomment-929788078

PR #875, opened 2021-10-21
Release GuitarChorder v2.0

flaneurette:
update

### DIFF
--- a/MIDI Editor/GuitarChorder.py
+++ b/MIDI Editor/GuitarChorder.py
@@ -1,0 +1,1672 @@
+# @description GuitarChorder
+# @author Flaneurette
+# @version 2.0
+# @changelog update
+# @provides [main=midi_editor,midi_inlineeditor] .
+
+try:
+    from reaper_python import *
+except:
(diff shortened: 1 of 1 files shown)


cfillion:
Works on Windows (out of the box after installing and enabling Python 3.8) and Linux (after configuring REAPER to load libpython3.9.so instead of the default libpython2.7.so + installing `tk`), but I'm not having luck getting this to run on macOS (10.14). Is this a known issue? (If so, the package can be set to only be available to install on Windows/Linux.)

After configuring REAPER to use Python 3.9 (instead of the default Apple-provided Python 2.7) and installing python-tk@3.9 via Homebrew, launching the script either does nothing, or hard-freezes REAPER. (Not familiar with Python or Tk, so not sure if there's something I can do to help debug this. `python3 -m tkinter` works  in a terminal and shows a window.)

flaneurette:
There have been a few bug reports on some versions of MAC. Would be best to set it to Windows/Linux, to be sure.

cfillion:
OK, done via f5ca90df7867e8e5cb65638e943061c4883f1477.

PR #883, opened 2021-10-26
BuyOne_Save-Load window set #1-10 with Mixer scroll position.lua

Buy-One:


Buy-One:
@cfillion 

Thank you, totally overlooked that bit. 

I'd like all the scripts to be installed inside one folder. For this to work should such folder pre-exist in the master repo or will it suffice to reference its name in the script names with a slash as they were listed initially? 
E.g. `[main] . > target_folder_name/script_name`

cfillion:
Yes, just specifying it in the provides line is enough. The directory doesn't need to exist on the repository.

PR #890, opened 2021-10-31
Release Split selected notes into new midi items v1.0

Rodilab:


cfillion:
A similar script exists in the repository: "Split selected MIDI item by notes" by Buy-One.

Rodilab:
Buy-One's script don't do the same. Maybe the name of my script is not clear enough.

Buy-One's script splits selected MIDI objects at each note it contains (horizontally)

My script "split vertically". This means : 
- Create a new empty MIDI item with same position and length as the selected item
- Copy selected notes and past them into new item
- Delete selected notes into original MIDI item

This script can be used, for example, to separate notes played with the left and right hand into two different objects.
Or separate de kick part after drum performance.

Rodilab:
Maybe the name should be "Split vertically selected notes from selected MIDI items"

Rodilab:
Or "Split vertically selected MIDI items with selected notes"

cfillion:
Or perhaps even "Move selected notes to a new overlapping item". Ahh... naming... 😨 

So, which name will it be?

EDIT: Since it use selected notes, would it be useful to have the action registered in the MIDI editor too? Seems like the kind of script that could be handy with a ME shortcut.

Buy-One:
How about "Move selected notes to a new MIDI item"?

Rodilab:
You are right. I like "Move selected notes to a new MIDI item", thanks :)
"would it be useful to have the action registered in the MIDI editor too?". Yes, you are right.

PR #897, opened 2021-11-08
Release Save project plugin info to text file (Don't open notepad on Windows) v1.06

Edgemeal:


cfillion:
The new action should be added to the original package instead of separately.

PR #906, opened 2021-11-18
Release Toggle snap relative to grid globally (Main and MIDI Editor) v1.0

JRTaylorMusic:


cfillion:
This script should be added to the other "Toggle snap globally" package because it's a variation on the same concept. That way both will be installed/updated together.

PR #913, opened 2021-11-23
Mordi MIDI Editor scripts - expand open URL function to Linux & ARM Mac

gablur:


cfillion:
Pinging @Landeplage (Mordi).

Also: 1) `open` with an empty string as first argument opens an extra Finder window 2) `CF_ShellExecute(url)` (from SWS) would be a more robust (and fully cross-platform) alternative to `os.execute`.

Landeplage:
Ah, neat. If we can avoid using SWS, I think that would be nice since some folks don't have it installed. Not sure what would be best. Feel free to merge anyway.

PR #931, opened 2021-12-31
Fix non-English text in trim mode (amagalma_gianfini_ReaNamer) 

wastee:
## Issuse

Lua standard library string's function `sub` did not fit with non-English text.

The length of non-English text is 3. And English or number's length is 1.

So, `string.sub` will cause issues with non-English text.

## Changed

> No change compared to original if text is English or number.

I write some code call `Helper for non-English string` to fit with non-English text.

Diff url: https://github.com/wastee/ReaTeam-ReaScripts/commit/cad2101d6202fe7a73ca09f1471ac15382e6937c

### test

Trim Start 3
Trim End 3

### result

Peek 2021-12-31 16-32

### DIFF
--- a/Various/amagalma_gianfini_ReaNamer (track-item-region renaming utility).lua
+++ b/Various/amagalma_gianfini_ReaNamer (track-item-region renaming utility).lua
@@ -99,6 +99,34 @@ function class(base, init)
    setmetatable(c, mt)
    return c
 end
+
+--/////////////////////////////////
+-- Helper for non-English string //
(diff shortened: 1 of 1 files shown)


cfillion:
This is a release repository, not a development repository. Patches should be sent to the author of the script, @amagalma in this case. https://forum.cockos.com/showthread.php?t=190534

Also, sounds like Lua's `utf8.offset` would be useful here. https://stackoverflow.com/a/43139063/796375

PR #940, opened 2022-01-19
Create BuyOne_Time selection - set start ; end point.lua (for MIDI Editors)

Buy-One:


cfillion:
The "BuyOne_" author prefix is missing from the filename. How about naming it "BuyOne_Set start and end point of time selection.lua"?

Scripts can't exactly match the naming of the Main section's native actions anyway.

Buy-One:
I didn't include my handle on purpose, because i didn't do anything special here other than try to adapt the native actions to the MIDI Editor context.

Do you mean in this format the script cannot be added to the action list?

Buy-One:
@cfillion 

I guess the requirements for merging are met now. Will this be merged or can i close the PR?

PR #954, opened 2022-02-02
Release Low latency monitoring v2.1

ak5k:
New streamlined and strongly opinionated version.

Same core functionality as before but with no user setting 'hassle'.

Assumes Per-chain PDC compensation mode (default in REAPER version > 6.19). Works with other modes too in a kind of 'will not break things' way.

Plugins can now be manually enabled/bypassed while Low latency monitoring script is running, and manual user choices are persistent. This makes previous including/excluding plugins or tracks with tags obsolete.

Hard limit is no longer available.  No separate PDC or individual plugin limit settings. PDC limit is fixed to one audio driver block/buffer. This should leave enough 'PDC headroom' to begin with. It's now just easier and faster to 'manually override' plugins anyway. 

Simpler and, hopefully, more robust.

### DIFF
--- a/Various/ak5k_Low latency monitoring.lua
+++ b/Various/ak5k_Low latency monitoring.lua
@@ -96,9 +96,7 @@ local function UpdateState()
   local retval
   retval, bsize = reaper.GetAudioDeviceInfo("BSIZE")
   bsize = tonumber(bsize) or 0
-  if limit == 0 then
-    limit = bsize
-  end
+  limit = bsize
   local masterTrack = reaper.GetMasterTrack(0)
   local automation = reaper.GetGlobalAutomationOverride()
   for i = 0, reaper.GetNumTracks() do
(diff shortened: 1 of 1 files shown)


Various/ak5k_Low latency monitoring.lua
```diff
@@ -1,7 +1,18 @@
 -- @description Low latency monitoring
 -- @author ak5k
--- @version 1.7
--- @changelog Bug fix related to track validation, which could cause script to crash when changing projects or deleting tracks while script is enabled.
+-- @version 0.2.1
```
cfillion: The new version number is lower than the previous one (v1.7 > v0.2.1). This won't trigger an update in ReaPack. Would using 2.1 instead be OK?

ak5k: I forgot one 'debug check' in. ~~Maybe I could close this pull request and re-upload with cleaned code and version v2.1?~~ Edited the pull request. v2.1 is ok.

PR #966, opened 2022-02-11
Update BuyOne_Lock FX and FX chains.lua

Buy-One:


cfillion:
The code change appears to be missing.

Buy-One:
This is a metapackage and I bumped the version because i bumped it in the underlying files. Should the version tag be removed from the metapackage entirely?

cfillion:
Ah, I didn't notice the other two had a noindex tag. They don't need any metadata header beside just noindex. Only the metapackage needs version information (etc.).

Please update all files belonging to a package in the same PR next time (preferably using the [package editor](https://reapack.com/upload)) to keep the git history clean, ensure all is merged in the right order, and make it easier to review.

Buy-One:
@cfillion  Noted

PR #972, opened 2022-02-14
added window position persistence for the select every N things scripts

benjohnson2001:
https://forum.cockos.com/showthread.php?t=241361


benjohnson2001:
Oh I actually need to add one more thing, please hold off on this merge

benjohnson2001:
okay it should be all set now

PR #1000, opened 2022-04-01
Release export track group names v1.0.0alpha

nicola-lunghi:
First release

### DIFF
--- a/Project Properties/nicks_export_track_group_names.lua
+++ b/Project Properties/nicks_export_track_group_names.lua
@@ -1,0 +1,69 @@
+-- @description export track group names
+-- @author NICKS
+-- @version 1.0.0alpha
+-- @changelog First release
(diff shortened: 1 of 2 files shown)


nicola-lunghi:
Hi thanks for checking this there's any issue?let me know its my first reaper extension :-)

cfillion:
Both Import and Export scripts should be part of a single package, rather than released individually. That way they'll always be installed and updated together. Click on the "Add File" button to add a file to the package in the editor.

The package's display name should use sentence case (eg. "Import/export track group names"). The action filenames too (no undescores except to separate the author prefix): "nicks_Import track group names.lua".

nicola-lunghi:
opened a new pr https://github.com/ReaTeam/ReaScripts/pull/1005

PR #1002, opened 2022-04-05
Release Add project markers as take markers into selected items v1.0

Landeplage:


X-Raym:
@cfillion this was duplicated of ** X-Raym_Add take markers from project markers to selected takes.lua ** .... but it has few bugs like no take rate support than mine doesn't have, and some oddities as `samplerate = reaper.GetSetProjectInfo(0, "RENDER_SRATE", 0, false)`which isn't used anywhere.
I wonder if it is useful to have this in the reateam repo then, as it is already exist in my repo ? What do you think ?

PR #1016, opened 2022-05-02
Release Scroll last touched track FX envelope into view v1.01

Edgemeal:


Edgemeal:
Any idea why this error happens?

Add a track fx param envelope using this script, don't add any points to the envelope, press the envelope panel bypass button so that REAPER hides the envelope (as expected), now call the script again and it errors,... BR_EnvAlloc' argument 1: expected TrackEnvelope* , but env is not nil (and still same userdata ), how can it not be valid?

cfillion:
What does `reaper.ValidatePtr(env, 'TrackEnvelope*')` returns?

Edgemeal:
false :(

Edgemeal:
I think I have it fixed, instead of checking env variable I used ValidatePtr()

THANKS!

PR #1020, opened 2022-05-21
Release Select floating FX track, unselect others v1.00

Edgemeal:


Edgemeal:
There is already a script like this one by me2beats, So just remove this!
**Auto select track with focused FX (defer)**

PR #1043, opened 2022-07-14
Release Select foldertracks based on track indentation v1.0

gus-lan:


X-Raym:
@gus-lan Hi ! There are two type of script header in this files, can you just keep the first one ?

X-Raym:
@gus-lan Hi, I just review the code of your script,
the main script has two headers, you should delete the second one and only keep the first.
The secondary script has also two header types, it is better to stick with one.

Another thing is that it seems that all the secondary scripts are in fact not using the main one code or functions to run, so you could rather share them as individual package rather than having one as a package for the others. 

But I see you have your own script repo: https://github.com/gus-lan/reaper_scripts
If you want to manage your own reapack sharing instead of using this repo (for faster release/updates), forking this repo template can help: https://github.com/cfillion/reapack-repository-template

Best

X-Raym (approved):
Better in one file.

PR #1051, opened 2022-08-22
Release MK Slicer v2.5

cool9-MK:
+ Added Pitch Detection mode for quick conversion of monophonic melodies and drums to MIDI.
+ The Drums preset in Pitch Detection recognizes and creates MIDI notes of Kick, Snare and Hat.
+ New overwrite mode for MIDI creation in Trigger and Pitch Detection modes.
+ Bug fixed: now the script does not crash when trying to process Empty Items.
+ Fixed a bug that slows down updating the script window when Loop is running.
+ Bug fixed: now after initializing items with a long silence at the beginning, Gain and Threshold do not take huge values.
+ Bug fixed: now if after initialization item selection disappears, all functions will still work.
+ Bug fixed: now Marker Quantize works if the visible grid is off.
+ Now when zooming horizontally with the keyboard keys, the waveform is centered on the edit cursor.
+ "Processing, wait..." message to brighten up your waiting while processing or initializing long items.
+ Service messages and caring error messages now appear at the top of the window.
+ The script runs even if the selection is not correct. No more intrusive pop-up windows. If unsuitable items are selected, the selection is removed automatically.
+ Now the script can run even on multitracks. Only the topmost selected track will be analyzed.
+ Now loop selection is not blocked and is captured by the script only at the moments of control from the Slicer interface.
+ The minimum value of the HPF slider turns off the HPF filter, completely removing filtering artifacts.
+ Limited minimum grid and ruler resolution: now long items and small grid resolutions do not overload the processor.
+ Optimization in general: now the script starts instantly. CPU load reduced by 50% in idle, memory consumption during processing and initialization is reduced.
+ Now, when initializing for multiple items, the script does not try to apply Heal by default. Glue only.
+ Smart Glue: On initialization, if a multitrack is selected, Glue happens selectively. Single items, MIDI and Empty Items are ignored.
+ Now in MIDI Trigger mode Glue will not work if the item's Rate is changed. The trigger has become non-destructive for single items.
+ Now the value of the pitch band is forced for ReaSamplomatic5000 instances. In the code, it is possible to turn off the boost or change the pitchband range.
+ After finishing MIDI processing, when no audio is loaded, loop selection is not captured by the script.
+ Improved responsiveness of sliders when using the mousewheel.
+ Increased the maximum width of the waveform window for ultra-wide screens.
+ Donate service changed to valid.

### DIFF
--- a/Items Editing/cool_MK Slicer.lua
+++ b/Items Editing/cool_MK Slicer.lua
@@ -1,15 +1,37 @@
 -- @description MK Slicer
 -- @author cool
--- @version 2.15
+-- @version 2.5
 -- @changelog
---   + Now the combination Ctrl+S works when the script window is active.
---   + os.execute replaced with CF_ShellExecute for better performance and Linux compatibility.
---   + Added a hidden option to force Sync to be enabled every time the script is run.
+--   + Added Pitch Detection mode for quick conversion of monophonic melodies and drums to MIDI.
+--   + The Drums preset in Pitch Detection recognizes and creates MIDI notes of Kick, Snare and Hat.
+--   + New overwrite mode for MIDI creation in Trigger and Pitch Detection modes.
+--   + Bug fixed: now the script does not crash when trying to process Empty Items.
+--   + Fixed a bug that slows down updating the script window when Loop is running.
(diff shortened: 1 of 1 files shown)


cfillion:
I've renamed the version number to 2.50 because 2.5 is smaller than 2.15.

PR #1092, opened 2022-12-03
Release Tempo marker list in time selection v1.0

Hypexed:


Buy-One:
Hi

The extension is Lua, but the language seems to be EEL2

Hypexed:
Ok this should be fixed as well now.

Buy-One:
@Hypexed

Also double hyphen used to comment out the header tags will produce error in EEL2

Hypexed:
@Buy-One

Thanks for info. Must have messed something up in package manager. It's getting exhausting updating on GH which I then duplicate in the package manager so definitely need a better work flow. :-)

Hypexed:
This should be good to go now.

cfillion:
Replaced by #1099.

PR #1100, opened 2023-01-02
Release FX Device v1.0beta

BryanChi:


cfillion:
1. The "Install file name" field is missing the .lua extension. You can leave this field empty to use the same "Storage file name" for both storage in the repository and installation on the user's computer.
2. Dependencies are missing. JSFX, fonts, images, etc that the script needs must be included in the package. Click on "Add file" in the package editor. For the JSFX files, be sure to set the "Resource type" to "JS effect".
3. The package editor will indicate where each of those dependencies will be installed relative to the resource path before uploading (below the "Install file name" field). Make sure the script is compatible with those new paths.

BryanChi:
Thanks for getting back so super quickly cfillion! 
I was actually about to ask on forum how to specify a path? When I enter / in the 'install file name' field, the would tell me the entered format is incorrect.
Also, I'm assuming paths can only be inside of 'ReaTeamScripts/category' right?

cfillion:
Slashes should be fine (eg. "foo/bar/baz.png"). (Maybe it was a backslash? Those are rejected.) What's the exact path you've tried that was detected as incorrect?

> Also, I'm assuming paths can only be inside of 'ReaTeamScripts/category' right?

The absolute technical limit is "ReaTeam Scripts". But as best practices, the main script file should be directly in the category subfolder, and all dependencies in a "author_Script name" sub-subfolder and beyond. (As per the default values.)

BryanChi:
Thank you! I've submitted another request

PR #1103, opened 2023-01-05
Update FXD Split To 4 Channels.jsfx

BryanChi:
Add 'FXD' as name prefix in description


cfillion:
Replaced by #1105 (merged both edit into a single PR + increased the version number (mandatory for ReaPack updates) + updated the changelog).

PR #1104, opened 2023-01-05
Update FXD ReSpectrum.jsfx

BryanChi:
Delete unused functions such as   draw_info_panel()    and     draw_mouse_crosshair()


cfillion:
Replaced by #1105.

PR #1106, opened 2023-01-05
Release FX Devices v1.0beta1

BryanChi:
-Remove calling deleted functions in ReSpectrum
-Update right names for Macros and Split to 4 channels and Main Script


cfillion:
Regarding the version number, I would recommend using v1.0beta1, v1.0beta1.1, v1.0beta2 etc, to allow a future v1.0 to happen. v1.01beta is a higher number than v1.0 so that would prevent a stable v1.0 from ever being released.

BryanChi:
Oh that's right, understood, is it possible for you to change it for me on your side..?

PR #1127, opened 2023-02-01
Release FX Devices v1.0beta9

BryanChi:
- New Major Feature: Band Split FX Container


FX/BryanChi_FX Devices.lua
```diff
@@ -41,13 +41,14 @@
 --   BryanChi_FX Devices/FX Layouts/ValhallaVintageVerb (Valhalla DSP, LLC).ini
 --   BryanChi_FX Devices/FX Layouts/ValhallaSupermassive (Valhalla DSP, LLC).ini
 --   BryanChi_FX Devices/FX Layouts/ValhallaDelay _Valhalla DSP_ LLC_.ini
+--   [effect] BryanChi_FX Devices/FXD Saike BandSplitter
+--   [effect] BryanChi_FX Devices/FXD Band Joiner
```
cfillion: These two don't have the standard .jsfx extension like the others. Should it be added (if so, is just renaming OK or the script needs an edit to match?) or OK like this?

BryanChi: Oh, it's working fine as is (without extension) on my Mac machine. But I think it's probably better to add it. I've tested adding extension name and it's working fine with the script.
I realized FXD (Mix)RackMixer is also lacking extension name. I'll add that too.

BryanChi:
Should I just re-upload from package editor?

PR #1152, opened 2023-03-01
Release FX Devices v1.0beta9.4-3

BryanChi:
-Fix Style Editor crash
-Fix console messages popping up
-Add Analog knob as a new style option


FX/BryanChi_FX Devices.lua
```diff
@@ -4500,13 +4532,13 @@
         function demo.PushStyle()
```
cfillion, line 4535: The error came from the addition of new SeparatorText style variables in 0.8.4. The style editor code in the demo isn't meant to be copied in other scripts so it doesn't handle well future vars. The commit responsible for updating the demo to Dear ImGui 1.89.3 is: https://github.com/cfillion/reaimgui/commit/5027834c0e3ddfd0271cc179605e71127e7fd38a#diff-f39b2c3b80f196dcc91c2638a61fad4452a3cf64aa09194dcdfe000dc7dae84eL6366. Specifically, it's the insertion of two new items in the `vec2` table in `GetStyleData`.

Scripts can simply include demo.lua and use its style editor directly. No need to duplicate the code:

```lua
local demo = dofile(reaper.GetResourcePath() .. '/Scripts/ReaTeam Extensions/API/ReaImGui_Demo.lua')
local ctx = reaper.ImGui_CreateContext('My script')
local function loop()
  demo.PushStyle(ctx)
  demo.ShowDemoWindow(ctx)
  if reaper.ImGui_Begin(ctx, 'Dear ImGui Style Editor') then
    demo.ShowStyleEditor(ctx)
    reaper.ImGui_End(ctx)
  end
  demo.PopStyle(ctx)
  reaper.defer(loop)
end
reaper.defer(loop)
```
It's meant as a temporary developer tool to help with styling rather than being a full user-facing configuration tool though.

BryanChi: Thanks for getting back so quick!
Looks like I was using old versions and that's why I wasn't experiencing any crashes.
I just updated and just by removing line 4469  -   r.ImGui_PopStyleVar(ctx, #cache['StyleVar']) , seems to have fixed it.

I do have some custom entries in the style editor so I guess I can't use the function as how you demonstrated?

cfillion:
> I do have some custom entries in the style editor so I guess I can't use the function as how you demonstrated?

Ah, right. Although commenting  out `PushStyleVar`/`PopStyleVar` makes the Alpha sliders and Border checkboxes do nothing. The proper fix would be updating the `vec2` table like in the linked commit. A better future-proof solution would be rewriting it to only push known style vars instead of blindly pushing all of the currently available ones (= not using `EachEnum`).

Also seeing these two lines in `PushStyle`:
```lua
if i == 54 then 
                                r.ImGui_PushStyleColor(ctx,54,get_aftr_Equal_Num(Lines[55]))
```

There is no guarantee 54 always has and will refer to the same color (as with any constant values in the ReaImGui API, very few are documented to be safe to hard-code). Right now `ImGui_Col_ModalWindowDimBg` is 54.

EDIT: Same with `ImGui_SetMouseCursor`. https://api.codetabs.com/v1/proxy/?quest=https://github.com/cfillion/reaimgui/releases/latest/download/reaper_imgui_doc.html#MouseCursor_Arrow
EDIT2: Noticed `SetMouseCursor`'s documentation did not mention that it takes `MouseCursor_*` values. [Adding that](https://github.com/cfillion/reaimgui/commit/d6bb8c8aa44e20a5737e0895db0695215516171d)!

BryanChi:
Thanks for reminding me about the SetMouseCursor, I'll fix that.

I'm still working on fixing the style editor. I think I will rewrite some parts myself because some parts are too advanced for me to understand  😵
May I please have your help on one thing though, how do I concatenate strings and then turn them into a function?
for example if I have **'r.ImGui_Col_'** and **'Button'**    and I want it to become this function -  **r.ImGui_Col_Button()**

I remember seeing this in the demo script...

cfillion:
```lua
local prefix = 'ImGui_Col_'
local suffix = 'Button'
local func = reaper[prefix .. suffix]
local color_id = func()

-- in short
local color_name = 'Button'
local color_id = reaper['ImGui_Col_' .. color_name]()
```

> I'm still working on fixing the style editor. 

Is v1.0beta9.4-3 good be released as-is, or should it wait until that's done?

BryanChi:
Oh right, thank you so much!

Yes the latest commit is good to be released!

PR #1158, opened 2023-03-09
Create BuyOne_Navigate to track send destination or receive source track via menu.lua

Buy-One:


X-Raym:
@Buy-One Hi, the way this script has to be manually edited before working is a bit unsual. As this is main repo for all user, can you change the system (having to put a character inside a string) so that

- it use boolean where appropriate in the user config area (simpler to understand, less subject to error)
- it actually works outside the box in with a certain default state (maybe All set to true) without having to manually edit it

thx!

Buy-One:
@X-Raym Hi Raymond, I do that on purpose so a user actually configures the script to work how they want it to, because there's a great chance that if they aren't presented with a prompt they will never look under the hood and won't discover some of its features they might find useful. And even if i mention the user settings in the About section they still might not read it.

And i ask very politely :)

However the script does work out of the box after ENABLE_SCRIPT variable is filled out.

X-Raym:
@Buy-One we can find less intimidating way to do this, like console message at first run (and save to ext state that first run occurs) :P
But IMHO the best is a core version of the script which runs without popup, and a second script which is just a GUI wrapper for setting config in extstate. Cause editing the script isn't future proof when there is reapack updates to make!

X-Raym:
At minima GetUserInput at first run where the user toggle things on/off if needed. Setting being saved in ExtState.

Buy-One:
> Cause editing the script isn't future proof when there is reapack updates to make!

I realize that update might ruin the settings, but this problem is not specific to my scripts. I personally don't like keeping extended states and using external files which is a dependency as far as i'm concerned, so tend to keep it as compact as possible.

X-Raym:
> but this problem is not specific to my scripts.
Precisely, but I can't review all scripts !

ExtState is perfect for such usage, it is just one line of code easyli accessible via API, it is really meant for that,

I understand you dont want to care with UI and settings, but IMHO at least working default and a popup would be nice,
and because it requires editing the script, just using the simple Preset Script trick to at least have one way for the user be mod securely (if he is ready to take the extra steps)

Take a look at this readme and scripts in this repo for more information about how let user have a future proof way to hook user settings
https://github.com/X-Raym/REAPER-ReaScripts/tree/master/Templates/Script%20Preset

what do you think about this?

Buy-One:
@X-Raym Yeah i considered the preset (settings) script and in principle i like the idea and for myself i'd probably use it, but again it creates a dependency for a user, and as i've learned many are not so script savvy to be able to manage dependencies, therefore a "one stop shop" approach seems to me the most user friendly even though it does have some downsides.

> at least working default and a popup would be nice

OK, i'll think about that

X-Raym:
Preset script is a dual system. it means that the user Can use preset script if it doesnt want his edits to be erased. But he can still edit the strict directly if he prefer.
From dev side, it is just putting the init call in a If statement

```lua
if not preset_file_init then
  Init()
end
```

Absolutely no downside here. Just a + for those who know the trick.

Buy-One:
@X-Raym I've changed the pop-up so it can be dismissed without editing the script. A working default was already there though. Empty settings were providing that.

PR #1160, opened 2023-03-11
Update BuyOne_Insert selected FX or FX chain presets in OR copy focused FX to selected objects.lua

Buy-One:


X-Raym:
@Buy-One hmmm I think you need version bump on the last commit as well ? not sure. Is this needed @cfillion ?

cfillion:
No need to bump multiple times in the same PR because all commits get squashed into a single one.

PR #1162, opened 2023-03-13
Release Script: Create 10 Minute Regions v1.0

alxgomes:


cfillion:
Scripts published here should be exclusively in English intended for an international audience. Also it's in the wrong category: should be Regions not Markers.

PR #1165, opened 2023-03-13
Release Create 10 Minute Regions v1.0 v1.0

alxgomes:


X-Raym:
Hi, I just reviewed the code of this script, it lacks a bit of polish for this repo,

- the name should be more explicit, like Create 10 minutes long regions from time 0 to end of project
- there is a weird loop inside the script, it seems it checks every first items of all tracks for each regions creation, which seems overkill, especially that it doesnt seems to stop iteration if a track below has an item too and can then call the Add Marker function several times instead of one
- There is a prefix in portugese for the region creation which I think will surprise lots of users, mayve just leave it empty as default.
- the last region doesnt last 10 minutes which would not be the most desired output IMHO
- it needs project start offset support

Question: Has this code been generated by an IA?

alxgomes:
the last region is at the end of the project, which can be less than 10 minutes

Yes, it was generated by an IA

how it worked for me tried to share

X-Raym:
@alxgomes thanks for sharing, the script do work but it needs few adjustment to be up to our standard. For eg right now it calls Create region" for each tracks having an items and for each regions, which isnt optimized and can even be problematic on big projects. Maybe I will fix this myself if you can't.

AI generated scripts are better on outside repositories, cause we better have here scripts which can be maintained by their authors (repos owner like me can't maintained all the scripts here alone).

Thanks anyway!

alxgomes:
> @alxgomes obrigado por compartilhar, o script funciona, mas precisa de poucos ajustes para estar à altura do nosso padrão. Por exemplo, agora ele chama Create region" para cada faixa que tem um item e para cada região, o que não é otimizado e pode até ser problemático em grandes projetos. Talvez eu mesmo conserte isso se você não puder.
> 
> Os scripts gerados pela IA são melhores em repositórios externos, porque é melhor termos aqui scripts que podem ser mantidos por seus autores (o proprietário de repositórios como eu não pode manter todos os scripts aqui sozinho).
> 
> Obrigado de qualquer maneira!

He doesn't create for each track, he creates for the project. Consider all tracks, as 3 for example may be the biggest. if you want to adjust it I appreciate it. I have no more knowledge for that.

X-Raym:
> He doesn't create for each track, he creates for the project.

I'm reading the code here.

https://github.com/ReaTeam/ReaScripts/pull/1165/commits/9e6b90ba96f16dbc1251dfb60de905d1c346e188#diff-5997bc52a82c7f6d1b46d711abcc8dad154905c9784c170cff944e0127308f54R36-R42

See: there is a track loop and a AddMarker function call inside this loop, which means the functions is called for every track (with an item in this case), despite only one call would be needed. If at the end only one region is created it is because of sanitization made on REAPER side, but propper data should be sent right from the start for cleaner script.

> if you want to adjust it I appreciate it.

Yeah this makes me thing this is why we cant really accept AI based stuff, it requires way less time for the author than from the code maintainers so it asks the question of "who is the author" of the script. We prefer authors to be able to take care of such fixes. I'll discuss this with other ReaTeam members.

alxgomes:
I apologize, I didn't mean to cause any inconvenience.

X-Raym:
@alxgomes oh dont worry for that, we are happy to see people involved in reascript and want to share things ^^
We just have incentive to have good quality and maintained code here cause it also serve as reference for other coders.

I may close this pull request, but thx for your participation!

alxgomes:
> @alxgomes oh não se preocupe com isso, estamos felizes em ver pessoas envolvidas em reascript e querem compartilhar coisas ^^ Nós só temos incentivo para ter boa qualidade e código mantido aqui porque ele também serve como referência para outros codificadores.
> 
> Posso encerrar este pull request, mas thx pela sua participação!

Yes you can, thanks for your work!

alxgomes:
feel free to take the idea and improve if you find it useful!

X-Raym:
Ok thx ! I close this pull request and let you know if I rebuild the script from scratch. Cheers!

PR #1187, opened 2023-04-07
Release FX Devices v1.0beta9.5.4-1

BryanChi:
- Fix mistaken FX position when drag FX from FX adder into BandSplit.


BryanChi:
Sorry for uploading so frequently, there was a small bug in the last version.

PR #1192, opened 2023-04-30
Release Pro Tools-like marker jump behavior v1.0.2

rexendevar:
fixed behavior with manually-numbered markers having higher numbers than the total count. no longer provides error message on failure.

### DIFF
--- a/Markers/rekseffeks_protoolsmarkerjump.lua
+++ b/Markers/rekseffeks_protoolsmarkerjump.lua
@@ -1,0 +1,19 @@
+-- @description Pro Tools-like marker jump behavior
+-- @author Rek's Effeks
+-- @version 1.0.2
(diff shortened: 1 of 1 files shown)


cfillion:
REAPER has this feature built-in: "View: Jump to time window" or double-clicking the time in the transport bar. The only difference is it needs the marker number to be prefixed by 'm' eg. `m1` or `m2`. Also supports jumping to a marker by name + includes regions and timestamp jumps.

rexendevar:
It is slower to use the built-in version as you have to type m before the marker number. This mean it can't be done entirely on the numpad with one hand as my script can. SWS extension has actions to launch ReaConsole with a letter already in the field to simplify the entering of data; jump to time window does NOT, or else I would just be using that.

PR #1194, opened 2023-05-21
Add files via upload

jkooks:
Add new Tree class to Lokasenna_GUI v2 classes


cfillion:
The version number of the Lokasenna GUI library v2 package needs to be increased to ship an update for it. However this is a release repository not a development one, so only completed new releases from or authorized by the package's owner can be accepted.

This could instead be released as an independent, separately installable package. https://reapack.com/upload/reascript

jkooks:
Alright, no worries, and thanks! I will see about adding it onto ReaPack.

PR #1197, opened 2023-05-31
Release Adjust selected tracks volume v1.0

l4bl:


cfillion:
SWS has "SWS: Open console with 'v' to trim volume on track(s)" which does a similar functionality.

l4bl:
> SWS has "SWS: Open console with 'v' to trim volume on track(s)" which does a similar functionality.

Oh you're right, sorry. I've not noticed that one. Should I close this request then right? Sorry I am a total noob on Github. Thanks in advance.

PR #1199, opened 2023-06-02
Release Set Loop from Time Selection v1.0

johnjallday:


cfillion:
There is already a native action that does this: "Time selection: Copy time selection to loop points".

PR #1207, opened 2023-07-06
Release FX Devices v1.0beta9.6.2fix1

BryanChi:
-Fix console message popping up when adding fx with enter key


BryanChi:
So sorry I forgot to remove one of my debugging thing in the last version!

cfillion:
Changed the version to "1.0beta9.6.2.1" because "1.0beta9.6.2fix1" is a pre-release for (= lower than) "1.0beta9.6.2".

PR #1217, opened 2023-07-19
Release FX Devices v1.0beta9.6.6

BryanChi:
-add more shortcut key options
-Fix 'When keys without modifiers have actions assigned to them, keys with modifires do not work. (e.g. If both Q and Shift + Q key have each action assigned to each of them, Q key action is triggered although I press Shift + Q key.)'


cfillion:
Are the `Msg` calls at line 799 and 824 leftovers?

Also could do this to obtain an array of all keys in one go:


```lua
local AllAvailableKeys = {}
for name, func in pairs(reaper) do
  name = name:match('^ImGui_Key_(.+)')
  if name then AllAvailableKeys[name] = func() end
end
```

BryanChi:
oh  yes the msg calls are unintentional, sorry I thought I checked!

Thanks! I remember there was something like this in the theme editor but don't remember how it was written. Oh well, I think I'll leave it as it is for now since there are some keys I want to exclude.

PR #1219, opened 2023-07-21
Release FX Devices v1.0beta9.6.6-2

BryanChi:
- fix unable to open script.... 2nd attempt...


BryanChi:
so sorry for multiple pull requests in one day, I tested earlier and it was working, but seems like I must have made a mistake somewhere!

PR #1222, opened 2023-07-25
Release Multitrack project creation v1.0.0beta

joabeslopes:


joabeslopes:
Instead, i think i prefer to add my own index.xml to the Reapack Repository list.

PR #1232, opened 2023-09-02
Release FXD - Vertical FX List v0.8

BryanChi:


Tracks Properties/bryanchi_FXD - Vertical FX List.lua
```diff
@@ -0,0 +1,1720 @@
...
+-- @about
+--   Vertical FX list that's similar to Doppelganger's fxlist script with some additional featrures:
+--   See forum post for detail: 
+--   https://forum.cockos.com/showthread.php?t=282282
+
+local r = reaper
+
+
+dofile(r.GetResourcePath() .. '/Scripts/ReaTeam Scripts/FX/BryanChi_FX Devices/Functions/General functions.lua')
+dofile(r.GetResourcePath() .. '/Scripts/ReaTeam Scripts/FX/BryanChi_FX Devices/Functions/FX Adder.lua')
```
cfillion, line 21: I'd suggest having this new script provide a separate copy of these files instead of requiring the user to install both.

Another advantage of that would be that FX List would always have a known working version of them while still allowing the copy in FX Devices to freely evolve in possible breaking ways.

This can be done automatically by ReaPack without you having to upload two copies of the files to the repository. Whenever a new version of FX List is released, it would take FX Devices' files at that point in time.

Would that be OK for you?

BryanChi: Right! I meant to change that but I forgot. Thanks for checking it for me!
I'll open a new pull request

PR #1234, opened 2023-09-03
Release FX Devices v1.0beta9.6.7

BryanChi:
FX Devices 9.6.6-4
- New : Shift+ RMB to bypass a modulation.
- Fix Slider default width shows as 0 , and inputting values by typing don’t work.
- Fix unable to RMB drag to adjust modulation range without choosing a source when there’s only one modulation on the parameter.
- Add Clap plugins support for FX adder !!! Thanks to Suzuki’s contribution !!!!


BryanChi:
Thank you so much for fixing the mistakes I made in the uploading process!  I think it's ready to be merged?

PR #1238, opened 2023-09-05
Release FX Devices v1.0beta9.6.7

BryanChi:
- Fix crash upon opening.
- Remove debug messages that's left unintentionally.


BryanChi:
So sorry for the frequent pull requests! There were some error that prevented it from opening.

PR #1240, opened 2023-09-09
Release FX Devices v1.0beta10.1

BryanChi:
- New feature!  attach draw list items on to a parameter - drawings can be affected by the respective parameter’s value by setting the ‘value affect’ property.
- Automatically copy image file into FX Device’s Image folder when user drag and drop image
- Style menu drop down menu : show all images as styles in the parameter type’s respective image folder.

- Fix Sequence window opening when mouse is adjusting other parameters but hovering over the modulator.
- Fix various unintended behavior when layout edit window is overlapping FX Device window.
- Fix Switch buttons default to 0xffffffff (fully white) color.
- Fix draw mode broken types - Circle, Circle filled, V-line, Text


FX/BryanChi_FX Devices.lua
```diff
@@ -1554,10 +1638,15 @@ function GeneralFunctions()
     end
 end
 
+function CopyFile(old_path, new_path)
+    local CopyCmd = r.GetOS():find("Win") and 'copy "' or 'cp "'
+    local ok = os.execute(CopyCmd .. old_path .. '" "' .. new_path .. '"')
```
cfillion: This code is dangerous: generating a command line from outside data to be executed by the system shell without proper sanitization.

(And also somewhat unreliable as copy and cp are outside executable that could do anything – there are no guarantees that `cp` actually 1) exists 2) is a program that copies files 3) is used that way.)

Copying files can be done without leaving Lua: https://forum.cockos.com/showthread.php?p=2360581.

cfillion (changes requested):

FX/BryanChi_FX Devices.lua
BryanChi: Thanks cfillion, I've updated the function with the one you provided, all seems to be working well!

PR #1250, opened 2023-09-16
Release FX Devices v1.0beta10.3.2

BryanChi:
- Add new FX browser that supports browsing FXs that works like the native FX browser. (Huge thanks to Sexan and Suzuki)
- Add repeat color for rectangles
- Add value affect for colors. ( Thanks for the idea Suzuki! )
- Renamed Draw mode into Background edit mode.
- Add Text font size for background edit mode.
- Increase font size range to 64.

---

Bug Fixes

- Fix adding FX Layering crashes sometimes
- Fix various behaviors regarding fonts and font sizes.


FX/BryanChi_FX Devices.lua
```diff
@@ -55,8 +63,6 @@
 --   Please check the forum post for info:
 --   https://forum.cockos.com/showthread.php?t=263622
 
-
-
 --------------------------==  declare Initial Variables & Functions  ------------------------
 VersionNumber = 'V1.0beta10.3.1 '
```
cfillion: Missing in-script version bump?
```suggestion
VersionNumber = 'V1.0beta10.3.2 '
```

BryanChi: Oops, yes,  thank you!

PR #1254, opened 2023-09-18
Release Snapshooter v1.1

tiagolr:
Transitions are now written to time selection when available
Added transition controls to preserve edges, invert transition, curve points tension and shape


tiagolr:
closing this, adding a small fix

PR #1256, opened 2023-09-28
Release Background Auto Save v0.9

BryanChi:
first release

### DIFF
--- a/Project Properties/bryanchi_Background Save.lua
+++ b/Project Properties/bryanchi_Background Save.lua
@@ -1,0 +1,53 @@
+-- @description Background Save
+-- @author Bryan Chi
+-- @version 0.9
+-- @changelog first release
+-- @about Uses Ultraschall background copy so Reaper doesn't freeze when saving a big project. Runs in a defer loop to always save the project's latest changes.
+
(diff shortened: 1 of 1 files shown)


X-Raym:
@BryanChi the way the loop is done here could be optimized:

https://github.com/ReaTeam/ReaScripts-Templates/blob/master/Templates/X-Raym_Background%20script%20-%20run%20action%20every%20X%20seconds.lua

Also, dont forget error message if ultraschall isnt present.

https://github.com/Ultraschall/ultraschall-lua-api-for-reaper#usage

It also miss some basic backround stuff like reporting toggle button state :

https://github.com/ReaTeam/ReaScripts-Templates/blob/master/Templates/X-Raym_Background%20script.lua

BryanChi:
Thanks for checking this, I've updated it. The other pull request is actually a different script, but I messed up the names when I uploaded and it's not letting me change them. This script should be 'Background auto save', which is a defering script, and the other one called 'Background Save', which is a one time save.

BryanChi:
I'm closing it because I just realized it actually doesn't work......   sorry!

PR #1263, opened 2023-10-01
Release MIDI Keyvel Filter v1.0

tiagolr:


tiagolr:
I submitted this JSFX for a bigger script called `Keyboard Splitter` that requires this FX, would be nice if both can be downloaded from reapack thanks

cfillion:
If this effect is meant solely for use along with the script and not standalone, then it can be included inside of the script's package itself.

Otherwise it can be submitted as an effect package on the [JSFX repository](https://github.com/ReaTeam/JSFX) rather than a script with a .lua file extension.

Either way it needs to be marked with the "JS effect" resource type to be installed in the corresponding directory in the user's resource directory (rather than "Scripts").

tiagolr:
Think that solves it, I'll cancel both PRs and submit the JS effect inside the same package

PR #1277, opened 2023-10-19
Release Track Color Layouts v1.0

BenTalagan:
Pull request for releasing the first version of a new tool, "Track Color Layouts".

The installation generates 4 children script for 4 layouts from the base "virtual" action script.


BenTalagan:
Note : I've patched my own PR to add 4 toolbar icons for color layouts. They are provided directly to the user in the toolbar_icons folder so that the toolbar integration is minimal (click, set, and go).

BenTalagan:
@cfillion, something wrong with this PR ? :-) (just asking because it was not merged like others yesterday).

BenTalagan:
Note : added one commit to this PR to update script metadata.

cfillion:
- The filenames and package name should use sentence case: "Track Color Layouts" -> "Track color layouts"
- The action filename suffixed with " action.lua" would be a bit odd since that's shown in the user's Action List
- The library file should be in the "talagan_Track color layout" directory (also sentence case)

BenTalagan:
> * The filenames and package name should use sentence case: "Track Color Layouts" -> "Track color layouts"

On this point I had a hard time between macos, reapack-index and git due to case sensitvity. I had to git push force my branch to avoid problems, because precedent commits in the PR would trigger warnings with reapack-index.

> * The action filename suffixed with " action.lua" would be a bit odd since that's shown in the user's Action List

Caution, it is **not** shown in the user's action list due to the rewriting of the action to different versions (layout 1, layout2, etc).

```
  [main=main,midi_editor] . > talagan_Track color layout 1.lua
  [main=main,midi_editor] . > talagan_Track color layout 2.lua
  [main=main,midi_editor] . > talagan_Track color layout 3.lua
  [main=main,midi_editor] . > talagan_Track color layout 4.lua
```

When proceeding that way, reapack skips the writing of the main file in the target dir. So the file is not even present in the REAPER script directory, and it's not a problem ; but in the meanwhile, to be cleaner, I've renamed the action file to, simply, "talagan_Track color layouts.lua" and have moved the lib file to the "talagan_Track color layouts" subdirectory. I have however kept the syntax "layout 1", "layout 2" (singular, not plural) for the actions because each one of those actions is linked to one and only one layout individually.

> * The library file should be in the "talagan_Track color layout" directory (also sentence case)

Done (except that I've used the "talagan_Track color layout**s**" directory).

I hope this is cleaner and good to be merged now, thank you for the pertinent remarks.

cfillion:
All good, merging!

> it is not shown in the user's action list due to the rewriting of the action to different versions (layout 1, layout2, etc).

Ah yes, renaming removes the implicit inclusion of the original filename.

BenTalagan:
Thanks, Christian !

PR #1281, opened 2023-11-05
Release Reduce interactively the number of events in Midi CC lanes v1.0.1

rgruet:
Doc changes

### DIFF
--- a/MIDI Editor/rig-trimCCs.eel
+++ b/MIDI Editor/rig-trimCCs.eel
@@ -1,0 +1,532 @@
+// @description Reduce interactively the number of events in Midi CC lanes
+// @author rig
+// @version 1.0.1
+// @changelog Doc changes
+// @provides [main=main] .
+// @about
+//   # rig-trimCCs
+//
+//    Reduce iteratively the number of CC events (points) in the selected MIDI editor lane.
+//
+//    Only CC lanes are supported (e.g. velocity or Channel pressure/aftertouch won't work).
+//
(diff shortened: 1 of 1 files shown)


MIDI Editor/rig-trimCCs.eel
```diff
@@ -0,0 +1,532 @@
...
+function init()
+(
+    gfx_init(SCRIPT_NAME, 230, 170);      // show script window
+    
+    // Attach "pin on top" button (if Reaper 6.24+ with JS extension installed):
+    // FIXME: works randomly when script launched from toolbar!!
+    //GetAppVersion(#reaper_version);
+    //strnicmp(#reaper_version, "6.24", 4) >=0  &&  APIExists("JS_Window_Find")  ? (
+    (APIExists("JS_Window_Find") && APIExists("JS_Window_AttachTopmostPin"))  ? (
+        JS_Window_AttachTopmostPin(JS_Window_Find(SCRIPT_NAME, true));
```
cfillion, line 510: Sorry for the long delay. This fails to parse if js_ReaScript API is not installed. This should use the `extension_api("JS_Window_...", ...)` syntax instead.

If the name of the script is just "Trim CCs", then the filename should be "rig_Trim CCs.eel" and `@description` should be "Trim CCs" to match.

The documentation mention adding the script to the MIDI Editor section, but the package is configured to add the script to the Main section only. Consider registering the script to the MIDI Editor section or both sections.

PR #1293, opened 2024-01-07
Release Mixing Station track name import v1.0

LostViking09:


Tracks/dontcupthemic_Mixing Station track name import.lua
```diff
@@ -0,0 +1,519 @@
...
+        else
+            reaper.MB("No track names found.", "Error", 0)
+        end
+    else
+        reaper.MB("No file selected.", "Error", 0)
+    end
+end
+
+-- Run the script
+reaper.defer(main)
```
cfillion, line 518: Why delay main until up to 30ms later? The entire contents of `main` could be probably extracted out of it and put directly in the script, no function or defer required.

Tracks/dontcupthemic_Mixing Station track name import.lua
```diff
@@ -0,0 +1,519 @@
...
+  if idx <= #str then
+    decode_error(str, idx, "trailing garbage")
+  end
+  return res
+end
+
+-- Function to prompt the user for a file
+function getUserFile()
+    local userFile = reaper.GetResourcePath()
+    local _, file = reaper.GetUserFileNameForRead(userFile, "Please load an MS Scene (Default: C:\\Users\\[user]\\MixingStaion\\Presets)", "json")
```
cfillion, line 409: Typo? `s/MixingStaion/MixingStation/`

What's the default path on macOS and Linux? If this script is only meant for Windows, installation should be limited to it using the "Target platform" field.

Shouldn't the MixingStation Presets folder be given to `GetUserFileNameForRead` instead of REAPER's resource path?

By the way Lua has `select` for skipping unused leading values.

PR #1296, opened 2024-01-16
Release Deselect odd-numbered items from selected items v1.0

milAgarr:


Items Editing/kkkanten_Deselect odd-numbered items from selected items.lua
```diff
@@ -0,0 +1,24 @@
...
+for i = 0, selectedItemsNum, 1 do
+  selectedItemsArray[#selectedItemsArray + 1] = reaper.GetSelectedMediaItem(0, i);
+end
+
+for i = 1, #selectedItemsArray, 2 do
+  if selectedItemsArray[i] == nil then
+    break
+  end
+  reaper.SetMediaItemSelected(selectedItemsArray[i], false);
+  reaper.ThemeLayout_RefreshAll();
```
cfillion: `ThemeLayout_RefreshAll` is the wrong function for this as it does more work than necessary for just refreshing the displayed item selection state. `UpdateArrange` is the function you're looking for.

Refreshing should be done once outside of the loop as there's no reason to slowly refresh the screen over and over.

`SetMediaItemSelected` does some internal state computation as well: `PreventUIRefresh` should be used to optimize those out when doing a loop.

The script could also benefit from having a named undo point rather than the default "ReaScript: Run".

It can be simplified into:
```lua
local UNDO_STATE_ITEMS = 4

reaper.Undo_BeginBlock()
reaper.PreventUIRefresh(1)
for i = reaper.CountSelectedMediaItems(nil) - 1 & ~1, 0, -2 do
  local item = reaper.GetSelectedMediaItem(nil, i)
  reaper.SetMediaItemSelected(item, false)
end
reaper.UpdateArrange()
reaper.PreventUIRefresh(-1)
reaper.Undo_EndBlock('Deselect odd-numbered items from selected items', UNDO_STATE_ITEMS)
```

Variants for even-numbered items and selecting might also be useful. In which case all four should be released together in a single "Select or deselect even or odd-numbered items" multi-action package.

milAgarr: Thank you for your reply and review!
I'm a beginner at creating scripts, so I learned a lot.

My question is, am I correct in understanding that the revised script should be submitted again from the ReaPack Package editor website for review?

I apologize for the inconvenience, but I would appreciate it if you could reply.
(Google translation)

cfillion: Yes, you can either close this PR + submit a new one or directly edit this one (via the "File changed" tab above -> "..." button at the right -> Edit this file).

No inconvenience at all, and welcome to ReaScripting! 😸

PR #1298, opened 2024-01-22
Release MK ReSampler v0.91

cool9-MK:
+ Added Fine slider for fine-tuning the frequency in cents.
+ Added Reverse function: play files backwards.
+ The waveform now displays Reverse if the function is active.
+ Now the script can receive MIDI from any channel, not just the first.
+ Fixed a bug: now play cursor correctly changes position if the zoom of the waveform is changed.
+ Fixed a bug: now the script does not give an error when working with Reversed items (but also does not work with them).
+ Fixed a bug: now the script does not crash with an error if the recording was activated from daw, but was stopped by the script button.
+ Fixed a bug: now the script does not crash with an error when pressing the keyboard arrows.
+ Now zoom, when you press the keyboard arrows, focuses approximately around the area of the start and end markers.
+ Added a condition for correct work with Muted items.
+ Improved display of the initial screen when launching on long items.
+ Now clicking on a sample in the Media Explorer resets the selection of items, shifting the focus of the script to the Media Explorer.
+ All menus that have only two items have been replaced with switch buttons with indication.
+ Increased the maximum deviation of Random Pitch to +-50 cents.
+ Changed the position of some controls.
+ Fixed typo: RealmGUI -> ReaimGUI.

### DIFF
--- a/Various/cool_MK ReSampler.lua
+++ b/Various/cool_MK ReSampler.lua
@@ -1,9 +1,26 @@
 -- @description MK ReSampler
 -- @author cool
--- @version 0.9
--- @changelog Initial
+-- @version 0.91
+-- @changelog
+--   + Added Fine slider for fine-tuning the frequency in cents.
+--   + Added Reverse function: play files backwards.
+--   + The waveform now displays Reverse if the function is active.
+--   + Now the script can receive MIDI from any channel, not just the first.
+--   + Fixed a bug: now play cursor correctly changes position if the zoom of the waveform is changed.
+--   + Fixed a bug: now the script does not give an error when working with Reversed items (but also does not work with them).
+--   + Fixed a bug: now the script does not crash with an error if the recording was activated from daw, but was stopped by the script button.
+--   + Fixed a bug: now the script does not crash with an error when pressing the keyboard arrows.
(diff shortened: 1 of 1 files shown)


Various/cool_MK ReSampler.lua
```diff
@@ -1,9 +1,26 @@
 -- @description MK ReSampler
 -- @author cool
--- @version 0.9
--- @changelog Initial
+-- @version 0.91
```
cfillion, line 3: Are you sure about 0.91? That's a lot higher than .9. Anything below 91 won't be allowed in future updates. Version numbers are not decimals.

Did you mean 0.9.1?

cool9-MK: 0.9.1 - looks OK. I will re-upload this script with the correct version number. Thank you.

Version numbers are hell for me. I always thought that these were just tenths and hundredths of a unit, as in normal counting in small steps. I assumed that 0.91 is greater than 0.90 by only 0.01.

cfillion: Version numbers are an infinite list of number or text segments separated by any non-alphanumeric character (eg. `2024-01-22.beta` is a fully valid version number). Each segments are compared individually and text segment (pre-release indicator) sort lower than numbers.

PR #1299, opened 2024-01-22
Release MK Shaper/Stutter v1.20

cool9-MK:
+ Added Razor Edit.
+ Now the script can manage any selected envelopes. If the envelope is not selected before running the script, the script will create a volume envelope by default.
+ The name of the active Envelope is now displayed in the script interface.
+ Added envelope selector for Items (Volume, Pan, Pitch).
+ Rise and Fall for the Pan envelope have been adjusted. The Gain control changes the range.
+ The range has been increased and the principle of operation of the Gain regulator has been changed.
+ Added "Soft Attack" option.
+ Significantly improved script performance when processing Selected Area.
+ Reset now correctly removes multitrack envelopes and selected areas.
+ True Stutter mode: When the Shape control is at maximum, the envelope takes on a rectangular shape.
+ Fixed a bug: sometimes in item envelope mode the points were not placed correctly if the item had a changed ratio.
+ Fixed a bug: now changing the Shift option does not cause the script to crash if this is done before pressing the Shape button.
+ Fixed a bug: flickering and incorrect position of item envelope points in Rise and Fall modes.
+ Fixed a bug: points were not placed correctly if the item was placed at the start of the project.
+ Fixed a bug: now the script does not crash when trying to process Empty Items.
+ Fixed a bug: now the point of the first transient is drawn correctly if it is not at the start of the item.
+ Fixed a bug that slowed down the update of the script window when Loop was running.
+ Fixed a bug: now the DestroyAudioAccessor function works correctly.
+ Now loop selection is captured by the script only when controlled from the script interface.
+ The minimum grid and ruler resolution has been limited: now long items and small grid resolutions do not overload the processor.
+ Optimization in general: now the script starts instantly. CPU load reduced by 50% when idle.
+ Controls have been streamlined for more intuitive operation.
+ Work without items: if there are no items in the selected area, the script will run in Grid and Track Envelope mode.
+ Now, when working with midi items and without items, the script runs almost instantly, without wasting resources on processing.
+ Added the option to manage all envelopes at once. When disabled, the script controls only the selected one.
+ Now, when zooming horizontally using the keyboard keys, the waveform is centered on the edit cursor.
+ The minimum HPF value turns off the filter, completely removing filtering artifacts.
+ Increased the maximum waveform window width and main window size limits for better display on large screens.
+ Theme: Added the option "Large Font Size" (Options - User Settings (Advanced)), which increases the font size when the window size is large.
+ Undo Points are now created during envelope generation.
+ Donate service has been changed to valid.

### DIFF
--- a/Envelopes/cool_MK ShaperStutter.lua
+++ b/Envelopes/cool_MK ShaperStutter.lua
@@ -1,14 +1,41 @@
 -- @description MK Shaper/Stutter
 -- @author cool
--- @version 1.01
+-- @version 1.20
 -- @changelog
---   + Fixed: now Shift works correctly if the tempo is different from 120.
---   + Now the combination Ctrl+S works when the script window is active.
---   + os.execute replaced with CF_ShellExecute for better performance and Linux compatibility.
---   + Added a hidden option to force Sync to be enabled every time the script is run.
+--   + Added Razor Edit.
+--   + Now the script can manage any selected envelopes. If the envelope is not selected before running the script, the script will create a volume envelope by default.
+--   + The name of the active Envelope is now displayed in the script interface.
+--   + Added envelope selector for Items (Volume, Pan, Pitch).
+--   + Rise and Fall for the Pan envelope have been adjusted. The Gain control changes the range.
+--   + The range has been increased and the principle of operation of the Gain regulator has been changed.
+--   + Added "Soft Attack" option.
+--   + Significantly improved script performance when processing Selected Area.
+--   + Reset now correctly removes multitrack envelopes and selected areas.
+--   + True Stutter mode: When the Shape control is at maximum, the envelope takes on a rectangular shape.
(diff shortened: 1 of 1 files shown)


Envelopes/cool_MK ShaperStutter.lua
```diff
@@ -1,14 +1,41 @@
 -- @description MK Shaper/Stutter
 -- @author cool
--- @version 1.01
+-- @version 1.20
```
cfillion, line 3: Did you mean 1.2 or 1.02 (both are the same number)? 1.20 ("twentieth minor update of major version 1") might be higher than you think.

cool9-MK: 1.20 is ok. I've been making a lot of big changes to the code, but haven't found the time to publish it. That's why it's such a big leap.
Thank you!

PR #1305, opened 2024-02-03
Release SampleSamurai v1.0

Develolop:
The help window now contains some further links


cfillion:
The version number must be increased (eg. to 1.0.1) for existing users to receive the update. Otherwise it's only available to new installs or re-installs. Or perhaps that's the intent here?

Develolop:
Ah okay, I will keep that in mind for the next time. This should be no problem for this one

PR #1306, opened 2024-02-05
Release Smart split items using mouse cursor context (also edit cursor, razor area and time selection) v3.0

AZpercussion:
Major update to version 3.0
- A lot of options stored in user config
- Options window
- Many new features and bug fixes


cfillion:
A small packaging nitpick: the options file "[Open options for AZ_Smart split items by mouse cursor.lua](https://github.com/ReaTeam/ReaScripts/pull/1306/files#diff-3f9de9226b880af0c3f7b695ba097d4f30ab44f9a04bd5e10d212c18877da5af)" has uppercase AZ while the script action has it lowercase.

(Also the option file should also start by az_, to match the action naming convention.)

Would renaming it to "az_Open options for az_Smart split items by mouse cursor.lua" be OK? Renaming the main package to uppercase is also possible, but that's the equivalent of deleting it + creating a new one.

AZpercussion:
Ah, okey! let it be "az_Open options for az_Smart split items by mouse cursor.lua"

PR #1307, opened 2024-02-05
One Small Step v0.1 (initial release)

BenTalagan:
This is a PR for releasing the very first version of a new tool, One Small Step, which is an alternative step input method for REAPER.

It uses the tracks FX input chains instead of the MIDI control path, and adds other input methods (sustain pedal validation, REAPER action validation).


MIDI Editor/talagan_OneSmallStep.lua
```diff
@@ -0,0 +1,395 @@
...
+    reaper.ShowMessageBox("Could not install One Small Step's helper FX on the track.\n\nIf you just installed One Small Step, please try to restart REAPER to let it refresh its JFSX repository.", "Oops !", 0);
+    return;
+  end
+
+  ui_loop();
+end
+
+function stop()
+  updateToolbarButtonState(0);
+  reaper.ImGui_DestroyContext(ctx);
```
cfillion, line 389: This function no-op currently and will removed in ReaImGui 0.9, I highly recommend enabling the backward-compatibility module (imgui.lua) to shield the script against such breaking changes.

MIDI Editor/talagan_OneSmallStep Change note len - 1.lua
```diff
@@ -0,0 +1,9 @@
+-- @noindex
+-- @author Ben 'Talagan' Babut
+-- @license MIT
+-- @description This is part of One Small Step
+
+package.path = debug.getinfo(1,"S").source:match[[^@?(.*[\/])[^\/]-$]] .."?.lua;".. package.path;
+
+local engine_lib = require "talagan_OneSmallStep/talagan_OneSmallStep Engine lib";
+engine_lib.setNoteLen("1");
```
cfillion: A nice trick you can do to avoid having to keep in sync multiple small files like this:

```
@provides
  author_Some source file.lua > author_Install name - 1.lua
  author_Some source file.lua > author_Install name - 2.lua
  author_Some source file.lua > author_Install name - Foo.lua
  author_Some source file.lua > author_Install name - Bar.lua
```

Then have that one source file detect at runtime what it's meant to be:
```lua
local SCRIPT_NAME <const> = select(2, reaper.get_action_context()):match("([^/\\_]+)%.lua$")
local SLOT <const> = tonumber(script_name:match("%- (%d+)"))
```

MIDI Editor/talagan_OneSmallStep/images/note_1.lua
```diff
@@ -0,0 +1,55 @@
+-- @noindex
+-- @author Ben 'Talagan' Babut
+-- @license MIT
+-- @description This is part of One Small Step
+
+return "\z
```
cfillion, line 6: Out of curiosity, why did you encode the images as Lua files instead raw PNG blobs/`CreateImage`?

The original intent behind `CreateImageFromMem` was to allow embedding small image data directly into the script's source, or packing multiple ones in a single file. Though I guess this works too. 😄 

There's also the option of creating an image map, one big PNG and only showing portion of it each time. That should be more efficient than multiple small images too.

MIDI Editor/talagan_OneSmallStep.lua
```diff
@@ -0,0 +1,395 @@
...
+  # Credits
+
+    This tool takes a lot of inspiration in tenfour's "tenfour-step" scripts. Epic hail to tenfour for opening the way !
+
+--]]
+
+-------------------------
+
+--[[
+# Ruby script to convert from png > lua to load binary img for ReaImGui
```
cfillion, line 97: There's also "Dump file to string literal" in ReaPack for that.

MIDI Editor/talagan_OneSmallStep/talagan_OneSmallStep Helper lib.lua
```diff
@@ -0,0 +1,126 @@
...
+-- @author Ben 'Talagan' Babut
+-- @license MIT
+-- @description This is part of One Small Step
+
+--[[
+-- Ruby script to generate note FX params
+def putn(note)
+  puts "slider#{3*note     + 10}<0,127,1>Note #{note} Pitch"
+  puts "slider#{3*note + 1 + 10}<0,15,1> Note #{note} Channel"
+  puts "slider#{3*note + 2 + 10}<0,127,1>Note #{note} Velocity"
```
cfillion: REAPER v6.74 added JSFX pre-processing so that JSFX code can generate itself.

https://www.reaper.fm/sdk/js/preproc.php

An example of it used to generate sliders: https://github.com/ReaTeam/JSFX/blob/99510175e9fa087eac446df3b66dedb5e264d4fa/Utility/mschnell_Inter%20Track%20Parameter%20Modulation.jsfx#L26.

cfillion (commented):
Packaging looks all good, I can release this as-is. 👍 

I've also posted a few general observations/suggestions that might be relevant below.

MIDI Editor/talagan_OneSmallStep/images/note_1.lua
BenTalagan: It required a bit of archeology to recall why I've used this technique 😆 - I wrote OSS recently, but took some parts from an older script. I'm not sure exactly, but it's probable that I simply got inspired by the demo because it is my main source for learning (more than the API) and I copied how it was done there. Having all those little icons as lua files reminded me how we embed XPM icons in Qt/C++ projects for example, and this is a pattern I'm used too. But yeah, I admit it looks funny there ! For this time I'll probably leave things as they are since this is working but I keep the idea of the big PNG for the next version.

MIDI Editor/talagan_OneSmallStep.lua
BenTalagan: Cool ! I didn't have the reflex to check in ReaPack for those little convenient tools. Thanks for the tip !

MIDI Editor/talagan_OneSmallStep/talagan_OneSmallStep Helper lib.lua
BenTalagan: Missed that one. This is insane 😄 I'm going to give it a try !

MIDI Editor/talagan_OneSmallStep Change note len - 1.lua
BenTalagan: Thanks ! i'll take a little time to correct my PR on this point.

MIDI Editor/talagan_OneSmallStep.lua
BenTalagan: The compatibility module is now enabled.

MIDI Editor/talagan_OneSmallStep Change note len - 1.lua
BenTalagan: Fixed for the various note lengths. (Edit : And also for the note len modifiers).

MIDI Editor/talagan_OneSmallStep/talagan_OneSmallStep Helper lib.lua
BenTalagan: Done, this opens up a new world ! So cool ! Thanks for pointing me to this worthy feature.

BenTalagan:
@cfillion , I think I've finished this pass of corrections. I let you have the last word !

PR #1309, opened 2024-02-08
Release Insert project marker under mouse or play cursor v1.0

AZpercussion:


cfillion:
Would it make sense to release the "and edit it" action in the same package as the regular action? Since they are closely related & likely to be updated together.

AZpercussion:
Yes, why not, I just forgot about this.  So let whole package name will be "Insert project marker under mouse or play cursor"
Thanks for your support!)

cfillion:
OK, done in #1311.

PR #1326, opened 2024-02-27
Update OneSmallStep v0.9.4 > v0.9.5

BenTalagan:
Update for One Small Step, v 0.9.5. Changelog : 

  - [Feature] Added insert/cursor mode (inserts + move things forward, deletes backwards)
  - [Feature] Added pedal repeater
  - [UI] Added status icons for modifier modes/keys (insert / backward / insert+backward)


MIDI Editor/talagan_OneSmallStep Insert Commit back.lua
cfillion, line 1: There's a lot of these sub-actions and it keeps growing, perhaps they could be moved to the talagan_OneSmallStep folder to keep the parent MIDI Editor clean?

Normally (as done when using [reapack.com/upload](https://reapack.com/upload)), all secondary files are in a subfolder.

BenTalagan: You're right, and it's probably going to grow more and more. The original idea was to have all callable actions in the MIDI Editor dir, and lib files in subfolders. But now the tool is getting more sophisticated than expected, so there's probably a need to re-organise things. I'll take the time (and extra care) to do it tomorrow, since there's one or two things to be careful (action IDs are going to change for example, and cross referenced paths too). I'll move every secondary action in an "actions" subfolder of the main subfolder.

BenTalagan:
Ok, this should be good now. !

PR #1334, opened 2024-03-10
Release Trim left, right or both item edges via mouse and razor v1.2

AZpercussion:
- options window
- fixed bug for grouped items on hidden tracks, now they are trimmed

### DIFF
--- a/Items Editing/az_Trim left, right or both item edges via mouse and razor.lua
+++ b/Items Editing/az_Trim left, right or both item edges via mouse and razor.lua
@@ -104,10 +104,12 @@ end
 ------------------------
 
 function OptionsWindow()
-  if reaper.APIExists( 'ImGui_GetVersion' ) ~= true then
+  local imgui_path = reaper.GetResourcePath() .. '/Scripts/ReaTeam Extensions/API/imgui.lua'
+  if not reaper.file_exists(imgui_path) then
     reaper.ShowMessageBox('Please, install ReaImGui from Reapack!', 'No Imgui library', 0)
     return
   end
+  dofile(imgui_path) '0.7'
   OptionsDefaults()
   GetExtStates()
   local fontSize = 17
--- a/Items Editing/az_Trim left, right or both item edges via mouse and razor.lua
+++ b/Items Editing/az_Trim left, right or both item edges via mouse and razor.lua
@@ -1,27 +1,393 @@
 -- @description Trim left, right or both item edges via mouse and razor
 -- @author AZ
--- @version 1.1
--- @changelog -bug fix: treat several selected items
--- @link Autor's page https://forum.cockos.com/member.php?u=135221
+-- @version 1.2
+-- @changelog
+--   - options window
+--   - fixed bug for grouped items on hidden tracks, now they are trimmed
+-- @provides [main] az_Trim left, right or both item edges via mouse and razor/az_Open options for az_Trim left, right or both item edges via mouse and razor.lua
+-- @link Forum thread https://forum.cockos.com/showthread.php?t=288069
+-- @donation Donate via PayPal https://www.paypal.me/AZsound
 -- @about
---   #Trim left, right or both item edges via mouse and razor
+--   # Trim left, right or both item edges via mouse and razor
 --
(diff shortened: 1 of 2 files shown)


Items Editing/az_Trim left, right or both item edges via mouse and razor.lua
```diff
@@ -1,29 +1,395 @@
...
+  end
+  font = reaper.ImGui_CreateFont(fontName, fontSize, reaper.ImGui_FontFlags_None()) -- Create the fonts you need
+  fontSep = reaper.ImGui_CreateFont(fontName, fontSize-2, reaper.ImGui_FontFlags_Italic())
+  if imgui_version_num >= 18910 then
+    reaper.ImGui_Attach(ctx, font)
+    reaper.ImGui_Attach(ctx, fontSep)
+  else
+    reaper.ImGui_AttachFont(ctx, font)
+    reaper.ImGui_AttachFont(ctx, fontSep)
+  end
```
cfillion, line 343: This is unreliable. The correct way is enabling ReaImGui's backward-compatibility module before using any of its functions:

```lua
local imgui_path = r.GetResourcePath() .. '/Scripts/ReaTeam Extensions/API/imgui.lua'
if not r.file_exists(imgui_path) then
  -- ...
  return
end

dofile(imgui_path) '0.8.7' -- version check + backward compat

-- from this point onward it's safe to use r.ImGui_* functions (only of v0.8.7 as requested above)
```

If you want to be compatible with ReaImGui v0.7: specify that as the version then use `AttachFont` and none of the newer functions or features.

AZpercussion: I don't want to stick to old version. It's just a workaround for the font  function, it works for some range of versions and I guess it's enough, because  most of users keep imgui up to date, thanks to Sexan))
Anyway, thanks for the tip, I'll look how it works.
And hmmm, I have a couple questions.
So, will your way safe for future? If user have installed newer version, which hasn't some of current functions, will script work anyway if i call an older version, or user will get a message to install specified version?

cfillion: > So, will your way safe for future?

Yes, that's the whole point of doing `dofile('imgui.lua')('request_version')` (and why it's important for a public release).

- If the current version < requested version = error message is shown prompting the user to update
- If the current version > requested version = the API is downgraded to match the version requested by the script

`reaper.ImGui_GetVersion` before (or without) `dofile('imgui.lua')` is not reliable because it's subject to breaking changes just like the rest of the API (eg. v0.7 changed the order of its return values).

cfillion: Screen Shot 2024-03-12 at 03 51 52

AZpercussion: > eg. v0.7 changed the order of its return values

Hmm, but my script works for version 0.7 as well.

Do I understand right that dofile('imgui.lua')  sticks  the script to a particular version and I need to choose main version for my script rather than add behavior for another versions?

cfillion: > Hmm, but my script works for version 0.7 as well.

I meant that `GetVersion` changed between 0.6 to 0.7, not 0.7 to 0.8, as an example of things that may break in the future when not using imgui.lua. Maybe it will change again in 5 years? (no such plans, but who knows)

> Do I understand right that dofile('imgui.lua') sticks the script to a particular version and I need to choose main version for my script rather than add behavior for another versions?

Correct. By picking a version and limiting your use to that version's API, imgui.lua can protect the script from future breaking changes in upstream dear imgui and/or ReaImGui itself.

For instance, 0.9 will remove `DestroyContext`. By using `dofile('imgui.lua')('0.8')`, your script will keep working as if nothing happened when 0.9 is released.

AZpercussion: Okay! Thanks a lot for the clarification, I' ll make a fix. EDIT: Done

> For example, 0.9 will remove DestroyContext

Oops..)

PR #1345, opened 2024-04-03
Update previous script and upload new script

tormyvancool:
I updated my scrip:

**TormyVanCool_AudioBooks_ChapterMarker.lua**

to the very last version and
I uploaded another script that supports the previous one:

**TormyVanCool_AudioBooks_ChapterMarkerUpdater**

My goal is: 

in case this way to collaborate to ReaPack works, I will transfer all my ReaScripts to your official repository.


cfillion:
1. [TormyVanCool_AudioBooks_ChapterMarker.lua](https://github.com/ReaTeam/ReaScripts/pull/1345/files#diff-0777a3f69f51d119a25545bd351176033cb56a3f66b123b5d2c4ce092718e41b) contains leading newlines: reapack-index stops reading after the first empty line outside of a tag
2. Scripts should use the "authorname_Package name in sentence case.lua" file naming convention, like "[tormyvancool_Chapter marker for audiobooks.lua](https://github.com/ReaTeam/ReaScripts/pull/1345/files#diff-ea3128813d3fc9a71405a8282c51566c37779af7eba6164e222352cbb507e26b)"
3. It seems like both [ChapterMarker.lua](https://github.com/ReaTeam/ReaScripts/pull/1345/files#diff-0777a3f69f51d119a25545bd351176033cb56a3f66b123b5d2c4ce092718e41b) and [ChapterMarkerUpdater.lua](https://github.com/ReaTeam/ReaScripts/pull/1345/files#diff-4b00b45fa01e6492050923a8334f4f058b3fc7b2e5ec2cf08f6754f9d1462288) are closely related: if so they should be released together in the same package, not two separate packages
3. I highly recommend using https://reapack.com/upload to create pull requests: it ensures all metadata is valid and makes the whole process generally easier

tormyvancool:
@cfillion  I corrected the filenames and also removed the 2 empty lines.
About packages I wrote on the Forum. because it's not clear at all.

PR #1360, opened 2024-04-22
Release Smart split items using mouse cursor context (also edit cursor, razor area and time selection) v3.20

AZpercussion:
- Added TCP context action
- fixed bug with empty takes
- sticked to imgui version 8.7.6 for stable work

### DIFF
--- a/Items Editing/az_Smart split items by mouse cursor.lua
+++ b/Items Editing/az_Smart split items by mouse cursor.lua
@@ -1,7 +1,10 @@
 -- @description Smart split items using mouse cursor context (also edit cursor, razor area and time selection)
 -- @author AZ
--- @version 3.15
--- @changelog - fixed one silly bug for new users
+-- @version 3.2
+-- @changelog
+--   - Added TCP context action
+--   - fixed bug with empty takes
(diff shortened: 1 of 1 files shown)


Items Editing/az_Smart split items by mouse cursor.lua
```diff
@@ -1,7 +1,10 @@
 -- @description Smart split items using mouse cursor context (also edit cursor, razor area and time selection)
 -- @author AZ
--- @version 3.15
--- @changelog - fixed one silly bug for new users
+-- @version 3.2
```
cfillion: Version numbers are not decimal numbers. 3.2 is a lot smaller than 3.15 (13 minor versions smaller) and so existing users would not receive the update.

AZpercussion: Ah, so will 3.20 work fine?

cfillion: Yes. Should I change it to 3.20 and merge?

AZpercussion: Yes, please)

PR #1362, opened 2024-04-25
ReaKS 1.1

Kabraxist:
I'll be branching per minor release.

### DIFF
--- a/MIDI Editor/ugurcanorcun_ReaKS.lua
+++ b/MIDI Editor/ugurcanorcun_ReaKS.lua
@@ -1,14 +1,14 @@
 -- @description ReaKS - Keyswitch Articulation Manager
 -- @author Ugurcan Orcun
--- @version 1.0
+-- @version 1.1
 -- @changelog 
---  - Big workflow overhaul: Newly added KS notes will fit themselves in the flow better.
---  - You can bypass smart insertion with Shift key
---  - Improved Theme color matching
---  - Simplified Settings
---  - Added tooltips and info to Settings view
---  - Articulations view will highlight the active KS
--- @link Forum Thread https://forum.cockos.com/showthread.php?t=288344
+--  - Added KS Note offset to help trigger Keyswitches just before the note.
+--  - Smashed many bugs related to Single-Line mode.
+--  - Removed unused config variables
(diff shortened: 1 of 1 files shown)


cfillion:
1.1 to 1.15 is a big increase (version numbers are *not* decimal numbers) and it'll prevent any future releases with a lower number. 1.1 < 1.1.5 < 1.2 < 1.15 < 1.20.

Is this OK? I would suggest 1.1.1 for major version 1, minor version 1, patch 1.

Kabraxist:
Riiiight, 1.1.5 makes more sense! :) 
Can you change it on your end or should I resubmit?

MIDI Editor/ugurcanorcun_ReaKS.lua
```diff
@@ -437,7 +453,7 @@ local function loop()
                         if ActivatedKS[i] ~= nil then ImGui.PushStyleColor(ctx, ImGui.Col_Button, EnumThemeColors.G) end
 
                         if ImGui.Button(ctx, articulation, 100) then
-                            local isShiftHeld = ImGui.IsKeyDown(ctx, ImGui.Key_LeftShift) or ImGui.IsKeyDown(ctx, ImGui.Key_RightShift)
+                            local isShiftHeld = ImGui.GetKeyMods(ctx) == 8192
```
cfillion: Numeric values are not stable nor part of the API contract for backward compatibility. Only the names (ImGui.Something) are.

`ImGui.Mod_Shift` is 8192 today, but it could be something else tomorrow. It was something else a few versions ago.

If you don't care if other modifiers (ctrl/alt/super) are pressed at the same time:
```suggestion
                            local isShiftHeld = ImGui.IsKeyDown(ctx, ImGui.Mod_Shift)
```

If it needs to be exactly Shift and no other modifier:
```suggestion
                            local isShiftHeld = ImGui.GetKeyMods(ctx) == ImGui.Mod_Shift
```

Kabraxist: So committing suggestions is a feature of GitHub, huh? Terrific! :D
Thanks for the correction :)

PR #1364, opened 2024-04-26
Release Perfect Timing! - Audio Quantizer v0.25

80icio:
- updated for Reaimgui 0.9

### DIFF
--- a/Items Editing/80icio_Perfect Timing! - Audio Quantizer.lua
+++ b/Items Editing/80icio_Perfect Timing! - Audio Quantizer.lua
@@ -1,16 +1,7 @@
 -- @description Perfect Timing! - Audio Quantizer
 -- @author 80icio
--- @version 0.24
--- @changelog
---   New Features
---
---   - New Visualizer zoom bounds reference  over tracks on Main window
---   - Trigger lines show full quarters with thicker lines and sub divisions with thinner lines
---   - Corrected "clipping" Transient attack filter
---
---   BugFix
---
---   - fixed wrong trackcount before quantizing if in -Edit Track- mode
+-- @version 0.25
+-- @changelog - updated for Reaimgui 0.9
 -- @link Forum thread https://forum.cockos.com/showthread.php?t=288964
 -- @about
 --   # PERFECT TIMING! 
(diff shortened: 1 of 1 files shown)


cfillion:
Skimmed through the code and found some things that could become issues in the future. (I can merge this PR as-is if you want, this just friendly advice not code review.)

```lua
r.ImGui_IsKeyDown( ctx, 16384 ) -- (and others like this)
```

The underlying numeric values of constants are not stable nor covered by backward compatibility (except for the very few ones that are documented as such). I don't know which `ImGui.Key_*` 16384 currently corresponds to, but it could be anything else in the future.

```lua
function help_widget(string, zero_parameter, storequery)
  if r.ImGui_IsItemHovered(ctx, r.ImGui_HoveredFlags_DelayNormal()) and Help_state == true then
    r.ImGui_BeginTooltip(ctx)
```

The return value of BeginTooltip should be checked (`if Help_state and ImGui.IsItemHovered() and ImGui.BeginTooltip() then`). By the way 0.9 added `ImGui.BeginItemTooltip` which as shortcut for IsItemHovered+BeginTooltip.

```lua
r.ImGui_BeginTable(ctx, 'total'
r.ImGui_BeginTable(ctx, 'settings',
r.ImGui_BeginTable(ctx, 'Visualizertable'
-- + others
```

Same. BeginTable may return false (eg. when clipped).

```lua
draw_list = r.ImGui_GetWindowDrawList(ctx)
```
Should be within Begin/End of the window (or child window) you want the drawlist of. Preferably in the same scope drawing is done. It's technically a ReaImGui bug that it works outside.

```lua
function OpenURL(url)
  local OS = r.GetOS()
  if OS == "OSX32" or OS == "OSX64" then
    os.execute('open "" "' .. url .. '"')
  else
    os.execute('start "" "' .. url .. '"')
  end
end
```
This opens an extra Finder window on macOS and doesn't work on Linux (+ would be vulnerable to command injection if `url` wasn't hard-coded). SWS has `CF_ShellExecute(url)` for this purpose.

80icio:
thanks for your help! you don't need to merge it , I'll fix it and make a new pull request

PR #1384, opened 2024-05-12
Release Spleeter: Split audio items into stems v1.0

rexendevar:
Created the scripts. Thank you to ReaTrak for providing a baseline.


rexendevar:
I selected "don't install" under the base script. As far as I'm concerned the three scripts are equally important and should have all been considered base scripts, but I didn't know how to do that so all three of them are secondary instead. I don't see anything in the base script to suggest that it took "don't install" as an option though.

rexendevar:
Would also appreciate if someone could test whether this works on MacOS. Then I could expand the OS targets.

rexendevar:
I have just noticed a mistake that makes the script generate garbage files. I don't know how to edit the merge request so I have to start all over. Lol

PR #1386, opened 2024-05-12
Release Spleeter: Split audio item into stems v1.1

rexendevar:
Added tentative Windows support. I have not tested it.

### DIFF
--- a/Items Editing/ReksEffeks_spleeter/Spleeter2.lua
+++ b/Items Editing/ReksEffeks_spleeter/Spleeter2.lua
@@ -1,0 +1,53 @@
+-- @noindex
+
+reaper.Undo_BeginBlock()
+
+function getPath(str)
+	if reaper.GetOS() == "Win32" or reaper.GetOS() == "Win64" then
+		return str:match("(.*[\\])")
+	else
+		return str:match("(.*[/])")
+	end
+end
+
+if reaper.GetOS() == "Win32" or reaper.GetOS() == "Win64" then
+	pt = "python -m "
+	else
+	pt = ""
+end
+
+sel_item = reaper.GetSelectedMediaItem( 0, 0 )
+
+if not sel_item then 
+	reaper.MB( "Select one audio item.", "Select Audio Item", 0 ) 
+	end_script = 1 
+end
+
+if sel_item then
+	reaper.Main_OnCommand( 41173, 0 )
+	item_pos =  reaper.GetMediaItemInfo_Value( sel_item, "D_POSITION" )     
+	item_take = reaper.GetMediaItemTake( sel_item, 0 )
+	source = reaper.GetMediaItemTake_Source( item_take )
+	filename = reaper.GetMediaSourceFileName(source, "")
+	filepath = filename
+	filename_no_ext = string.gsub(filename, '[.]%w%w%w$','')
+	take_type = reaper.TakeIsMIDI( item_take )
+	if take_type then
+		reaper.MB( "Item is MIDI", "Select Audio Item", 0 )
+		end_script = 1
+	end 
+end
+
+if not end_script then
+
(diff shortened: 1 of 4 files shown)


rexendevar:
I don't know how to edit a pull request that's already in progress. Lol

rexendevar:
it does not work on windows

Items Editing/ReksEffeks_spleeter/Spleeter2.lua
```diff
@@ -0,0 +1,53 @@
...
+		end_script = 1
+	end 
+end
+
+if not end_script then
+
+	local cmd = (pt .. 'spleeter separate -p spleeter:2stems-16kHz -o \"' .. getPath(filepath) .. '\" \"' .. filename .. '\"')
+  
+	if reaper.MB( "       Splitting selected audio into 2 stems.\n\n                 Vocals / Accompaniment\n\nProgram will hang until task is done. Continue?", "Spleeter confirmation", 1 ) == 1 then
+		os.execute(cmd)
```
cfillion, line 46: This has a command injection vunerability. `filename` is arbitrary data and `cmd` is a shell script. While I don't expect it would be exploited maliciously, this bug can realistically cause the command to fail or behave unexpectedly.

`reaper.ExecProcess` is slightly safer because it only executes the desired program with the given arguments without interpreting the command line as a shell script, unlike `os.execute`. Unexpected inputs would still fail though. External command execution from ReaScript is inherently flawed with all APIs currently available.

Even if the above was fixed, I'm not sure it'd be a good idea to release this because it depends on both spleeter and python to be installed outside of ReaPack, so these script don't "just work" out of the box. There is no error handling either to guide users on what they need to do in case these system dependencies cannot be found.

cfillion (changes requested):

Items Editing/ReksEffeks_spleeter/Spleeter2.lua
rexendevar: Aside from error handling in the case of Spleeter not being present, is there anything that can be done to fix the issue? I previously tried running it with `reaper.ExecProcess` and the output audio files were simply not functional (see [results of an earlier demo](https://youtu.be/B_TbDpqCfc0)).

rexendevar: I do believe that it is extremely useful to provide a way to run Spleeter integrated into the DAW, however. I realize that it is not plug-and-play as most Reaper scripts are but I believe that running a short script from within the DAW is vastly superior to any manual method and therefore is worth the extra dependency.

cfillion: Perhaps this could be a Python ReaScript? Python has better process spawning APIs than Lua + ExecProcess. You'd need an API where the program name and its arguments are passed as separate strings, rather than a command line string with injected arbitrary data. (Python has many bugs in REAPER, but it might be fine for this use case.)

```py
program   = 'spleeter'
arguments = ['separate', '-p', 'spleeter:2stems-16kHz', '-o', getPath(filepath), filename]
```

On macOS and Linux such an arguments array goes straight to the process' `main` function without command line splitting/shell scripting involved. On Windows they are joined using [specific rules](https://daviddeley.com/autohotkey/parameters/parameters.htm#WINCRULESDOC).

"Program will hang until task is done" isn't great UX: it's frustrating for users and they may force quit REAPER if it takes too long, potentially losing unsaved data... I'm not familiar with Python, but it probably has ways to spawn and monitor background processes? If it does, could do a little UI with a cancel button eg.

Example

And since spleeter is itself written in Python, maybe it has some API you could use and avoid the external command execution altogether?

> I do believe that it is extremely useful to provide a way to run Spleeter integrated into the DAW, however. I realize that it is not plug-and-play as most Reaper scripts are but I believe that running a short script from within the DAW is vastly superior to any manual method and therefore is worth the extra dependency.

I'm sure it could be useful, but such a script should be quite a bit more fleshed out... eg. currently it displays "Done" and inserts 0s-long items if the command fails to generate the expected .wav files for any reason...

PR #1392, opened 2024-05-17
Release Zuljeah v0.1beta1

hemisphera:
First release


hemisphera:
Just wondering ... it's been sitting here for 1 month now. Is there anything that I'm missing for this to move on?

Various/hemisphera_Zuljeah/Zuljeah_Monitor.lua
```diff
@@ -0,0 +1,153 @@
...
+  for i = 0, itemCount - 1 do
+    local mi = reaper.GetTrackMediaItem(track, i)
+    local ok, take = reaper.GetSetMediaItemInfo_String(mi, "P_NOTES", "", false)
+    local pos = reaper.GetMediaItemInfo_Value(mi, 'D_POSITION')
+    local seq = tonumber(tostring(take))
+    for k,v in pairs(regions) do
+      if (v["start"] <= pos) and (v["end"] >= pos) then
+        v["mediaitem"] = mi
+        v["selected"] = reaper.IsMediaItemSelected(mi)
+        sortedRegions[seq] = v
```
cfillion, line 55: Sorry for the long delay... Tried it out: it crashes on this line (seq is nil) if `P_NOTES` fails to parse as a number.

The 'Consolas' font is only pre-installed on Windows.

The user-facing actions should have the author prefix and be named as sentence case (eg. "hemisphera_Zuljeah - Next song.lua") to blend well with other actions.

Since all three Lua files share a lot of code, I recommend putting the common code in one file and loading that from the others instead of duplicating. A good place would be the hemisphera_Zuljeah.lua package file (in which case it needs to be marked for installation, but not listed in the action list).

Is this behavior normal when starting playback after moving the cursor a bit before the regions?
Take

Docker #0 may be anywhere (top/bottom/left/right/floating), so it would be best to persist in which docker the user wants to put the window.

PR #1393, opened 2024-05-18
version 0.2 of kaddaok_Import UVR separations into comping lane structure.lua

KaddaOK:
- Now includes files in folders below the original FLAC's path as well as beside it.
 - Item name parsing now compatible with stems downloaded from X-Minus
 - Friendly item name for Mel-Roformer model from May 2024 UVR patch

### DIFF
--- a/Tracks/kaddaok_Import UVR separations into comping lane structure.lua
+++ b/Tracks/kaddaok_Import UVR separations into comping lane structure.lua
@@ -1,17 +1,27 @@
--- @description Import UVR separations into comping lane structure
+-- @description Import UVR or X-Minus separations into comping lane structure
 -- @author Kadda OK
--- @version 0.1
+-- @version 0.2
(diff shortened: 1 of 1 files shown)


Tracks/kaddaok_Import UVR separations into comping lane structure.lua
```diff
@@ -1,17 +1,27 @@
--- @description Import UVR separations into comping lane structure
+-- @description Import UVR or X-Minus separations into comping lane structure
```
KaddaOK, line 1: Is updating this `@description` an issue?  What will that mean for anyone currently using the script; does it still associate as being out of date?

cfillion: No, it's fine. `@description` is only used for display and searching. As long as the file is not renamed, there is no side-effects.

PR #1400, opened 2024-06-06
Release Mini song switcher v1.1

mschnell1:
Updated Description to correctly show underscores in markdown


cfillion:
@[about](https://github.com/cfillion/reapack-index/wiki/Packaging-Documentation#package-tags) is a package tag, so it can be changed without increasing the version.

PR #1403, opened 2024-06-19
Release Tempo Envelope - select env point under edit cursor v1.0

AudioBabbL:


cfillion:
This skips the first point and does an extra lookup past the last point. Also the "(deselect others)" action should be provided in the same package, instead of two separate packages.

AudioBabbL:
Thanks for pointing that out cfillion -- I will have to have a closer look, it's been a while!

PR #1412, opened 2024-07-02
Release Mini song switcher v2.1.1

cfillion:
Fix Lua comment in Mini song switcher project.eel


mschnell1:
Thanks !
I still am not clear what exactly I did did wrong with the upload tool. 
It suggested `lua` as file extension and I modified this to `eel`, thinking that would be the way to use it....

Maybe the pipeline check might detect such errors...

cfillion:
Screen Shot 2024-07-02 at 14 17 46

mschnell1:
I See  ! :) 
Thanks

PR #1419, opened 2024-08-04
Release Fade tool (works on context of mouse, razor or time selection) v2.11

AZpercussion:
repair saving of text field options


cfillion:
2.1 to .11 is a huge bump. Should it be 2.1.1?

AZpercussion:
> 2.1 to .11 is a huge bump. Should it be 2.1.1?

If it let me make the next version 2.2 - yes, 2.1.1 will better, else it's doesn't matter.

cfillion:
> If it let me make the next version 2.2 - yes, 2.1.1 will better, else it's doesn't matter.

Yes. 2.11 would have required the next version to be 2.12 or higher (or 2.11.X).

PR #1426, opened 2024-08-25
Release Floating FX bypass toggle (toggle action using last saved parameters) v1.1

80icio:


cfillion:
This action should be added to the "Floating FX bypass" package instead of a separate one just for it.

80icio:
> This action should be added to the "Floating FX bypass" package instead of a separate one just for it.

is there a way I can add it now or should i just make a new version and pull a request as a package?

cfillion:
Easiest would be opening a new pull request updating "Floating FX bypass".

PR #1431, opened 2024-09-16
Release Track Icon Selector v1.0

Alejandro-Reapertips:


Buy-One:
:thumbsup: Should have been a native feature similar to toolbar icon selector

Tracks/reapertips_Track-Icon-Selector.lua
```diff
@@ -0,0 +1,539 @@
...
+    if not r.ImGui_GetVersion then
+        deps[#deps + 1] = '"Dear Imgui"'
+    end
+    if #deps ~= 0 then
+        r.ShowMessageBox("Script needs ReaImGui.\nPlease Install it in next window", "MISSING DEPENDENCIES", 0)
+        r.ReaPack_BrowsePackages(table.concat(deps, " OR "))
+        return true
+    end
+end
+if NeedDeps() then return end
```
cfillion: The script will crash prior to reaching here, at the `ImGui_GetBuiltinPath`/ `require` lines above, if ReaImGui is not installed. The check should be done before that, and preferably be looking for `ImGui_GetBuiltinPath` (the only function guaranteed to always exist before enabling backward compat in "0.9.1" mode).

```lua
if not reaper.ImGui_GetBuiltinPath then
  return reaper.MB('ReaImGui is not installed or too old.', 'Track icon selector, 0)
end

package.path = reaper.ImGui_GetBuiltinPath() .. '/?.lua'
local imgui = require 'imgui' '0.9.2'
```

Tracks/reapertips_Track-Icon-Selector.lua
```diff
@@ -0,0 +1,539 @@
...
+local MAIN_PNG_TBL = {
+    [0] = {}
+}
+local FILTERED_PNG
+
+local icon_size = 32
+
+r.set_action_options(1)
+
+local ctx = imgui.CreateContext('Track Icons', imgui.ConfigFlags_DockingEnable)
```
cfillion: `DockingEnable` overrides the user's preference. Is this the desired behavior? Docking is enabled by default.

Tracks/reapertips_Track-Icon-Selector.lua
```diff
@@ -0,0 +1,539 @@
...
+--
+--   REAPER Folder > Data > track_icons
+--   Within this folder, create or modify folders to represent your desired categories. The script will automatically recognize these folders as categories in the sidebar, allowing you to filter icons based on their assigned categories.
+
+-- @description Track Icon Selector
+-- @author Sexan
+-- @license GPL v3
+-- @version 1.0
+-- @changelog
+--  Initial release
```
cfillion: Duplicate metadata header (I'll remove it before merging).

cfillion (commented):
Script should use sentence case to match REAPER actions and other scripts. I'll rename the file to `reapertips_Track icon selector.lua` and change the display name to `Track icon selector` before merging.

Tracks/reapertips_Track-Icon-Selector.lua
GoranKovac: yeah that flag can be removed

Tracks/reapertips_Track-Icon-Selector.lua
GoranKovac: my bad, added that code fast yesterday and did not looked where I've added it.

Alejandro-Reapertips:
Thank you so much @cfillion 🙏

PR #1440, opened 2024-10-03
Track icon selector pcall CreateImage

GoranKovac:
When unsupported image format is encountered show its full path in reaper Console


cfillion:
It might be useful to report the original error message from ReaImGui too, as there's a myriad of other failure modes beside "unsupported format". (Image too big, memory alloc failure, any error reported by the PNG/JPEG decoders...)

PR #1441, opened 2024-10-13
Release Song switcher (for live use) v1.7

cfillion:
run action markers from take markers in the current song's primary track (REAPER v6+)

### DIFF
--- a/Various/cfillion_Song switcher.lua
+++ b/Various/cfillion_Song switcher.lua
@@ -1,9 +1,7 @@
 -- @description Song switcher (for live use)
 -- @author cfillion
--- @version 1.6.1
--- @changelog
---   disable keyboard input when the filter text box or a context menu has focus
---   fix the toolbar buttons showing 1 frame early, before the filter text box is hidden
+-- @version 1.7
+-- @changelog run action markers from take markers in the current song's primary track (REAPER v6+)
 -- @provides
 --   [main] cfillion_Song switcher/cfillion_Song switcher - Send signal.lua > cfillion_Song switcher/cfillion_Song switcher - Switch to next song.lua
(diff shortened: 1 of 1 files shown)


mschnell1:
You might have noticed that I rewrote some parts of the song switcher (in eel, as I don't understand  LUA ) to allow for having it work in a background project. Works fine, but of course just a single "song switcher" project would be better. Maybe you might consider adding that functionality to the "main" song switcher protect,

cfillion:
Do you mean like in #1442? Fixes song validation not supporting background tabs (= the song turns to red even though the track still exists).

The new take action marker functionality does support background tabs.

mschnell1:
I am not sure.
What I intended with "Mini Song Switcher" is to have the Songs be hosted in a dedicated background project tab and be controlled by Midi activity in the foreground Tab. Hence when "calling" a song you need to give not only the track (I did this in the same way as "song switcher") but additionally the (potentially background) Tab to be accessed (i.e. started / stopped),while the "running" state of the foreground tab is not  affected . This can be done by ...._X()  API calls.

cfillion:
With that bugfix the "Switch to xyz" actions work no matter what is the current active tab. As long as the songs are themselves all in the same project.

(If there are multiple project tabs with songs instead of one, changes can be done by combining with the "Reset data" action to reload the song list. Change active tab, reset song switcher data, switch song, return to previous tab. Can be wrapped in a script with `PreventUIRefresh`.)

mschnell1:
The primary "problem" I found at that time was not about switching but about starting / stopping. 
The "Song Switcher" started the Tab in foreground, while the "Mini Song Switcher" can be configured to start / stop a project tab defined by the user. (Issue with my setup: the foreground Tab is used to host the instruments played by Midi Keyboard. )

Solution: using the ....Ex() API, which allow to specify the Tab to be addressed. (e.g. "Main_OnCommandEx")

(Dynamically switching the active Tap and then back when a Midi Message is received seems not like a really good way.)

PR #1445, opened 2024-10-23
Release Play lanes in time selection v1.00

Edgemeal:


Edgemeal:
I try to update this script to v1.01, but I got error on reapack.com when I click the pull request button.
**Error: There are merge conflicts**

cfillion:
Your fork's master branch is [4 commits ahead and 2 commits behind](https://github.com/Edgemeal/ReaScripts-1) upstream so the upload tool can't bring it up-to-date. Reset it so that it matches [the history](https://github.com/ReaTeam/ReaScripts/commits/master/) of this repo (or just delete/re-create it).

Edgemeal:
Thanks, did the Discard 4 commits option,  reapack uploaded with no errors, hope that worked!

cfillion:
Yup, all good. By the way, a slightly more robust check for ReaImGui's absence would be `if not reaper.ImGui_GetBuiltinPath then` since that's the only function meant to be used *before* enabling backward compatibility.

It's extremely unlikely, but let's say in 10 years Begin gets renamed to BeginWindow. The script would normally still work thanks to the backward compat, however it would then fail the presence check.

Whereas `ImGui_GetBuiltinPath` must not ever be removed because that would break every script.

Edgemeal:
Thanks, Good to know! 
Will change it for next update!

PR #1453, opened 2024-11-03
Release Scroll TCP to track currently edited in active MIDI editor v0.5

BenTalagan:
Simple script to scroll the TCP so that the currently active MIDI editor's take's track appears at the top of the arrange view.


cfillion:
Wouldn't the MIDI Editor category be a better fit for this than Various?

BenTalagan:
Good question ! I've been hesitating for a while on this ; at first I had put it into the "View" category like it's done in @X-Raym's repository before realizing this categorization was specific to him. Since it scrolls the arrange view, it's something hybrid between the MIDI editor and the arrange view, but because it has no effect when there's no open MIDI editor, it's probably better categorized in the MIDI Editor section. Changing this ...

BenTalagan:
Ok, done !

X-Raym:
@BenTalagan MIDI Editor is the only reapacl specific name folder to determine context.

What matters here is not on what it is triggered but from where. I imagine this action on MIDI editor toolbar so better put it there indeed.

BenTalagan:
@X-Raym I don't think it would not necessarily be triggered from the MIDI editor toolbar (could be in more generic actions that handle the current global layout), and it's just technically a small "brick" anyway (I don't use it alone), so even if there's a good chance that it appears in a custom action launched from the MIDI toolbar, I've still added it to both action sections [main, midieditor], but you're right, the probabilities are largely in favor of the MIDI Editor, so let's go this way ! It's already done anyway :-)

BenTalagan:
@X-Raym @cfillion, since we had this discussion here, I thought it was the right place to ask. I'm preparing a small pack of scripts that allows to minimize/maximize/change height of docks, currently those actions would be :  

- minimize dock that contains current midi editor
- maximize dock that contains current midi editor
- change height ' ' ' ' ' to (500)   <----- (could be clonable with another value in the title)
- minimize bottommost dock
- maximize bottommost dock
- change height ' ' ' ' ' to (500)   <----- (could be clonable with another value in the title)

What's your opinion on the right category to chose this time ? "Various" ? Maybe a category that does not exist yet ?

X-Raym:
@BenTalagan in my Free Repo, I would put in View category

https://github.com/X-Raym/REAPER-ReaScripts/tree/master/View

BenTalagan:
Thanks @X-Raym ! Seems to me perfectly pertinent, let's read @cfillion's opinion now since he has the last word :-)

PR #1457, opened 2024-11-06
Release YOUTUBE Downloader v2.9

tormyvancool:
1.0 2024-26-10
    # First Release
1.1 2024-26-10
    + Processes Notifications
    - /Video/
    + /Videos/
1.2 2024-26-10
    - --merge-output-format mp4
    + -S vcodec:h264,res,acodec:aac
1.3 2024-26-10
    - 10
    + 2
1.4 2024-26-10
    - 2
    + 5
1.5 2024-26-10
    - 5
    + 1
    # Unified Update
1.6 2024-26-10 - 1
    + 2
    + Version
1.7 2024-27-10
    - 'start "" "' from all O.S.s
    + 'start "UPDATE & DOWNLOAD" "' Win
1.8 2024-27-10
    - GGGGG = ''
    - 1
    + Start = '"'
    + 2
1.9 2024-27-10
    + Check saved project
    - 1
    + 2
2.0 2024-27-10
    - "chmod +x " ..  MainPath
    + 'chmod +x "' ..  MainPath .. '"'
    # Ordered Variables
    - 2
    + 1
    + Apple Trial
2.3 2024-27-10
    # Linux execution correction
    + Credits
    # 2.1 and 2.2 just trials due issues with Linux and Apple
2.31 2024-28-10
    # Binaries directly form the source
2.32 2024-28-10
    - yt-dlp
    + yt-dlp_linux
2.4 2024-29-10
    # Adjusted header style for production
2.5 2024-11-04
    - Various 
    + VideoPath = 'Video'
2.6 2024-11-05
    + check for temrination of temporary file upfrotn import the video
2.7 2024-11-05
    - Check Routine
2.8 2024-11-06
    + Detects Nework Interruptions during download
    + Removes leftovers
    + URLs as filename: forbidden
    + Limitation to only alphanumerical characters
2.9 2024-11-06
    + Check IfFileExists: Overwrite, Newname, Exit

### DIFF
--- a/Video/tormyvancool_YOUTUBE Downloader.lua
+++ b/Video/tormyvancool_YOUTUBE Downloader.lua
@@ -161,7 +161,7 @@ local CallPath = ScriptPath .. 'yt-dlp/' -- Get FullPath to yt-dlp
         local a = {}
         local MainPath = ''
         if OS == "Win32" or OS == "Win64" then
-          MainPath = '"' .. ResourcePATH .. '/Scripts/Tormy Van Cool ReaPack Scripts/' .. VideoPath .. '/yt-dlp/' .. dlpWin .. '"'
+          MainPath = '"' ..CallPath .. dlpWin .. '"'
           Start = 'start /b /wait "UPDATE & DOWNLOAD" '
           OpSys = 1
         end
@@ -172,17 +172,15 @@ local CallPath = ScriptPath .. 'yt-dlp/' -- Get FullPath to yt-dlp
           OpSys = 2
         end
         if OS == "Other" then
-         -- MainPath = ResourcePATH .. '/Scripts/Tormy Van Cool ReaPack Scripts/Various/yt-dlp/' .. dlpLnx .. '"'
-         -- Start = '"'
-         -- os.execute('chmod +x "' ..  MainPath .. '"')
-          MainPath = '"' .. ResourcePATH .. '/Scripts/Tormy Van Cool ReaPack Scripts/' .. VideoPath .. '/yt-dlp/' .. dlpLnx .. '"'
+          MainPath = '"' ..CallPath .. dlpLnx .. '"'
           Start = ''
           os.execute('chmod +x ' ..  MainPath)
           OpSys = 3
         end
         return MainPath
       end
-      
+
+
       -- GET FILE SIZE
       function get_file_size(filename)
           local file = io.open(filename, "rb")
--- a/Video/tormyvancool_YOUTUBE Downloader.lua
+++ b/Video/tormyvancool_YOUTUBE Downloader.lua
@@ -67,8 +67,9 @@
 --     + URLs as filename: forbidden
 --     + Limitation to only alphanumerical characters
 -- 2.9 2024-11-06
---     + Check IfFileExists: Overwrite, Newname, Exit
---     + check if the subdir for yt-dlp exists. if not it warns the user and stops the script
+--     + Checks IfFileExists: Overwrite, Newname, Exit
+--     + Checks if the subdir for yt-dlp exists. if not it warns the user and stops the script
+--     + If the video wasn't downloaded or the donloaded file is 0 bytes size, the script stops and doesn't generate any new track
 -- @about:
 -- # Import VIDEOs directly in TimeLine from YouTUBE, VIMEO, PATREONS and thousand other ones.
 --  
@@ -139,7 +140,7 @@ local CallPath = ScriptPath .. 'yt-dlp/' -- Get FullPath to yt-dlp
         a = a + 1
       until(CheckForDir == nil)
       if returnedDir == nil then
-        local retQuery = reaper.MB("This script must be isntalled from a Reapack Repository.\n\nClick \"OK\", remove it and install it as it should!\n", "INSTALLATION ERROR", 0)
+        local retQuery = reaper.MB('This script must be isntalled from a Reapack Repository.\n\nClick \"OK\", remove it and install it as it should!\n', 'INSTALLATION ERROR', 0)
         if retQuery == 1 then
           goto done 
         end
@@ -220,13 +221,13 @@ local CallPath = ScriptPath .. 'yt-dlp/' -- Get FullPath to yt-dlp
 
       -- CHECK WHETHER PROJECT IS SAVED
       if pj_name_ == "" then 
-        reaper.MB("YOU MUST SAVE THE PROJECT FIRST! Then relaunch this script!",'WARNING',0)
+        reaper.MB('YOU MUST SAVE THE PROJECT FIRST! Then relaunch this script!','WARNING',0)
         return
       end
 
       -- GET URL
       repeat
-      retval, url=reaper.GetUserInputs("DOWNLOAD VIDEO", 1, "Paste URL,extrawidth=400", InputVariable)
+      retval, url=reaper.GetUserInputs('DOWNLOAD VIDEO: URL', 1, "Paste URL,extrawidth=400", InputVariable)
       if retval==false then return end
       if retval then
         t = {}
@@ -248,7 +249,7 @@ local CallPath = ScriptPath .. 'yt-dlp/' -- Get FullPath to yt-dlp
       -- GET FILENAME
       ::getfilename::
       repeat
-      retval_1, FileName=reaper.GetUserInputs("DOWNLOAD VIDEO", 1, "Insert FILE NAME,extrawidth=400", InputVariable)
+      retval_1, FileName=reaper.GetUserInputs('DOWNLOAD VIDEO:TITLE', 1, "Insert FILE NAME,extrawidth=400", InputVariable)
       FileName = GetRid(GetRid(GetRid(GetRid(GetRid(FileName, pipe), colon), quote), slash), backslash) -- No reserved characters can be written
       FileName = FileName:gsub("http", "")
 
@@ -264,12 +265,12 @@ local CallPath = ScriptPath .. 'yt-dlp/' -- Get FullPath to yt-dlp
       
       -- NO EMPTY TITLE ADMITTED
       if t[1]== "" then
-        reaper.MB("VIDEO TITLE is MANDATORY","ERROR",0,0)
+        reaper.MB('VIDEO TITLE is MANDATORY','ERROR',0,0)
       end
       
       -- NO URLS AND NOT ALPHANUMERICAL CHARACTERS ADMITTED
       if t[1]:match("[^%w%s]") then
-        reaper.MB("ONLY ALPHANUMERIC CHARACTERS ADMITTED","ERROR",0,0)
+        reaper.MB('ONLY ALPHANUMERIC CHARACTERS ADMITTED','ERROR',0,0)
         t[1]=""
       end
       until( t[1] ~= "")
@@ -314,7 +315,7 @@ local CallPath = ScriptPath .. 'yt-dlp/' -- Get FullPath to yt-dlp
           local checkfile = reaper.file_exists(Destination)
           local answer = nil
           if checkfile == true then
-            answer = reaper.MB("A file with the same filename exists\nWould you want to overwrite it?\n\nYES => Go on\nNO => Rewrite the filname\nCANCEL => Exit",'WARNING: FILENAME EXISTS',3)
+            answer = reaper.MB('A file with the same filename exists\nWould you want to overwrite it?\n\nYES => Go on\nNO => Rewrite the filname\nCANCEL => Exit','WARNING: FILENAME EXISTS',3)
           end
           if answer == 7 then 
             goto getfilename 
@@ -349,12 +350,19 @@ local CallPath = ScriptPath .. 'yt-dlp/' -- Get FullPath to yt-dlp
             a = a + 1
           until(test == nil)
           if returned ~= nil then
-            local retQuery = reaper.MB("Due a Network Error, the video was not properly downloaded.\nBY CLICKING OK THESE LEFTOVERS WILL BE REMOVED\n\nLeftovers:\n\n" .. returned, "NETWORK ERROR", 0)
+            local retQuery = reaper.MB('Due a Network Error, the video was not properly downloaded.\nBY CLICKING OK THESE LEFTOVERS WILL BE REMOVED\n\nLeftovers:\n\n' .. returned, 'NETWORK ERROR', 0)
             if retQuery == 1 then
             os.remove(ProjDir .. "/Videos/" .. returned)
             end
           else
-            reaper.InsertMedia(Destination, 1)
+            if not reaper.file_exists(Destination) or get_file_size(Destination) == 0 then
+              local retQuery = reaper.MB('Something went wrong during the download of this video.\nNo video were downloaded.\n\nBy clicking on \"OK\" the script will stop.', 'AN ERROR OCCURRED', 0)
+              if retQuery == 1 then
+              goto done
+              end
+            else
+              reaper.InsertMedia(Destination, 1)
+            end
           end
 
           if debug == true then 
--- a/Video/tormyvancool_YOUTUBE Downloader.lua
+++ b/Video/tormyvancool_YOUTUBE Downloader.lua
@@ -125,7 +125,7 @@ local ProjDir = reaper.GetProjectPathEx(0)
 local ResourcePATH = reaper.GetResourcePath()
 local VideoPath = 'Video'
 _,ScriptName = reaper.get_action_context()
-local ScriptPath = ScriptName:match('^.+[\\/]') -- Script Path
+local ScriptPath = ScriptName:match('^.+' .. package.config:sub(1, 1)) -- Script Path
 local CallPath = ScriptPath .. 'yt-dlp/' -- Get FullPath to yt-dlp
 
       -- CHECK IF YT-DLP FOLDER IS CREATED BY REAPACK AS EXPECTED
@@ -148,7 +148,8 @@ local CallPath = ScriptPath .. 'yt-dlp/' -- Get FullPath to yt-dlp
 ---------------------------------------------
 -- FUNCTIONS
 ---------------------------------------------
-      
+
+
       -- SLEEP(SECONDS)
       function sleep(n)
         local t0 = clock()
--- a/Video/tormyvancool_YOUTUBE Downloader.lua
+++ b/Video/tormyvancool_YOUTUBE Downloader.lua
@@ -68,6 +68,7 @@
 --     + Limitation to only alphanumerical characters
 -- 2.9 2024-11-06
 --     + Check IfFileExists: Overwrite, Newname, Exit
+--     + check if the subdir for yt-dlp exists. if not it warns the user and stops the script
 -- @about:
 -- # Import VIDEOs directly in TimeLine from YouTUBE, VIMEO, PATREONS and thousand other ones.
 --  
@@ -85,11 +86,11 @@
 --    and several other ones ...
 -- 
 --   [Full list here](https://github.com/yt-dlp/yt-dlp/blob/master/supportedsites.md)
---
--- Credits:
+-- @Credits:
 --    Stefano marcantoni and Ben Talagan - to have helped for MAC implementation
 --    Paolo Saggese PMS67 - to have helped for Linux implementation
 --    cfillion - for his support during general debug
+--    smandrap - for his key suggestoin to that helped to improve the reliability. Topic https://forum.cockos.com/showthread.php?t=96087
 -- @provides
 --   [win64] yt-dlp/yt-dlp.exe https://github.com/yt-dlp/yt-dlp/releases/latest/download/yt-dlp.exe
 --   [linux] yt-dlp/yt-dlp_linux https://github.com/yt-dlp/yt-dlp/releases/latest/download/yt-dlp
@@ -123,8 +124,27 @@ local pj_name_ = reaper.GetProjectName(0, "")
 local ProjDir = reaper.GetProjectPathEx(0)
 local ResourcePATH = reaper.GetResourcePath()
 local VideoPath = 'Video'
-local CallPath = ResourcePATH .. '/Scripts/Tormy Van Cool ReaPack Scripts/' .. VideoPath .. '/yt-dlp/' -- Get FullPath to yt-dlp
+_,ScriptName = reaper.get_action_context()
+local ScriptPath = ScriptName:match('^.+[\\/]') -- Script Path
+local CallPath = ScriptPath .. 'yt-dlp/' -- Get FullPath to yt-dlp
 
+      -- CHECK IF YT-DLP FOLDER IS CREATED BY REAPACK AS EXPECTED
+      a=0
+      CheckForDir = ""
+      repeat
+        if CheckForDir == "yt-dlp" then
+          returnedDir = CheckForDir
+        end
+        CheckForDir = reaper.EnumerateSubdirectories( ScriptPath, a)
+        a = a + 1
+      until(CheckForDir == nil)
+      if returnedDir == nil then
+        local retQuery = reaper.MB("This script must be isntalled from a Reapack Repository.\n\nClick \"OK\", remove it and install it as it should!\n", "INSTALLATION ERROR", 0)
+        if retQuery == 1 then
+          goto done 
+        end
+      end
+      
 ---------------------------------------------
 -- FUNCTIONS
 ---------------------------------------------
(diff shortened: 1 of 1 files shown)


Video/tormyvancool_YOUTUBE Downloader.lua
```diff
@@ -0,0 +1,367 @@
...
+          ---------------------------------------------
+          
+          -- GET RESIDUAL FILES
+          local ResFiles
+          if OpSys == 2 or OpSys == 3 then
+              ResFiles = "ls -a " .. ProjDir .."/Videos/*.mp4.part"
+              ResDel = 'rm "' .. ProjDir .. '\\Videos\\*.mp4.part"'
+          elseif OpSys == 1 then
+              ResFiles = 'dir /b "' .. ProjDir .. '\\Videos\\*.mp4.part"'
+              ResDel = 'del "' .. ProjDir .. '\\Videos\\*.mp4.part"'
```
cfillion: There are API functions for this that don't require going through system commands. Also (at least) the Unix code path is not only dangerous but also broken: the star within quotes is interpreted literally not as a globbing operator. And no backslashes as directory separator on Unix.

Video/tormyvancool_YOUTUBE Downloader.lua
```diff
@@ -0,0 +1,367 @@
...
+--       + Detects Nework Interruptions during download
+--       + Removes leftovers
+--       + URLs as filename: forbidden
+--       + Limitation to only alphanumerical characters
+--   2.9 2024-11-06
+--       + Check IfFileExists: Overwrite, Newname, Exit
+-- @provides
+--   [windows] yt-dlp/yt-dlp.exe https://github.com/yt-dlp/yt-dlp/releases/latest/download/yt-dlp.exe
+--   [linux] yt-dlp/yt-dlp_linux https://github.com/yt-dlp/yt-dlp/releases/latest/download/yt-dlp
+--   [darwin] yt-dlp/yt-dlp_macos https://github.com/yt-dlp/yt-dlp/releases/latest/download/yt-dlp_macos
```
cfillion: Should un-list for unsupported 32-bit Windows/macOS, and for the script itself too so that it cannot be installed where it can't be used.

Video/tormyvancool_YOUTUBE Downloader.lua
```diff
@@ -0,0 +1,367 @@
...
+        local MainPath = ''
+        if OS == "Win32" or OS == "Win64" then
+          MainPath = '"' .. ResourcePATH .. '/Scripts/Tormy Van Cool ReaPack Scripts/' .. VideoPath .. '/yt-dlp/' .. dlpWin .. '"'
+          Start = 'start /b /wait "UPDATE & DOWNLOAD" '
+          OpSys = 1
+        end
+        if OS == "OSX32" or OS == "OSX64" or OS == "macOS-arm64" then
+          MainPath  = './yt-dlp_macos'
+          Start = 'cd "' .. CallPath .. '" && chmod +x ' .. dlpMac .. ' && '
+          os.execute('chmod +x "' ..  MainPath .. '"')
```
cfillion, line 173: Same issue as last time here.

cfillion (commented):
It's an improvement over last time however there are still significant design flaws that I can't accept on a default repository. Most notably the complete absence of any protection against injection which is unchanged from the previous PR.

In the current state I think it would be better released on your own repository.

Video/tormyvancool_YOUTUBE Downloader.lua
tormyvancool: I tried that command in Linux virtual machine (Ubuntu) as it is, before writing it into this  code and it works.
About the APIs you are referring to: which are?
Are they doing exactly what I want from these lines. The goal is to remove any `.mp4.part` file that's forming inside that specific subfolder.
If knew they existed I could use them; APIs are making my life easier (when and IF correctly explained)

NOTE "any protection against injection"???

Video/tormyvancool_YOUTUBE Downloader.lua
tormyvancool: For the unsupported 32bits I will modify what's written between `[]` but how to avoid to download the script itself on 32bit machines: not any clue (I never did it).

EDIT/ on the ReaPack website I see now the platform dropdown is valid also for the script;
Should I just add the  ` -- @provides [win64] ` oly?
tu there is not an option for also MAC and Linux ... I f I chose oen I can't chose for the other ones as well.

How to make it downloadable only from 64bit users for ALL O.S.es?

Video/tormyvancool_YOUTUBE Downloader.lua
tormyvancool: **If you are talking about the line 145**: it seems in Mac when you call directly the application (1 single command instead of all that "CDs") it returns that the file doesn't exists.

**If you are talking about line146**: On Linux I tried to give more than one time `chmod +x` everything works. So I don't see evidence to do it in a different way since when the yt-dlp is downloaded it seems do not have the execution flags. Only readable.

Video/tormyvancool_YOUTUBE Downloader.lua
cfillion: > and it works.

It cannot for the two reasons listed above. If the temp files are removed it's by yt-dlp not by those two commands.

Screen Shot 2024-11-08 at 05 47 37

> about the APIs you are referring to: which are?

`reaper.EnumerateFiles` and `os.remove`

> "any protection against injection"???

From the previous PR:

> Sure the macOS/Linux code paths (double-quoted w/o escaping) likely works for most cases. But it will fail for some users because the full range of valid inputs is not considered. And the failure mode (arbitrary command injection) allows for catastrophic outcomes (even if unlikely in practice).
>
> Input is any non-null bytes representing a path to an existing file. Output is a shell script to be interpreted by sh, bash or cmd.exe. All inputs must map to a valid output.
>
> Right now many inputs can make the shell behave unexpectedly (technically [arbitrary command injection](https://owasp.org/www-community/attacks/Command_Injection)).
>
> (cmd.exe has very different syntax rules than sh/bash! APIs such as os.execute are hard to get right and [widely](https://www.reddit.com/r/C_Programming/comments/uokyqw/comment/i8f5ot1/) [discouraged](https://wiki.sei.cmu.edu/confluence/pages/viewpage.action?pageId=87152177).)

Video/tormyvancool_YOUTUBE Downloader.lua
cfillion: Repeat the `provides` line for the script itself for each platform (dot = shortcut for the current file):

```lua
-- @provides
--   [win64]        .
--   [darwin64]     .
--   [darwin-arm64] .
--   [linux]        .
```

Using the package editor: click on the duplicate file button in the list on the left.

EDIT: Ah the editor does not detect this idiom and instead generates more verbose metadata (`[linux] . > author_Filename.lua`). It's functionally equivalent, just redundant (`>` is the rename operator). Will look into improving that...

Video/tormyvancool_YOUTUBE Downloader.lua
cfillion: Line 146 only runs on macOS. It attempts to set the execute bit on yt-dlp_macos in whichever directory is current (unlikely to be yours the first time). So the line is effectively no-op. Only the one at line 145 is effective (ignoring the same injection problem). By the way it's checking `OS == "OSX32"` despite the script (well, the yt-dlp binary) not supporting macOS 32-bit?

Video/tormyvancool_YOUTUBE Downloader.lua
tormyvancool: yt-dlp does not remove these files self when the network fails. No way.
These commands are entering in actions ONLY if at the end of the process, yt-dlp failed/crashed and .mp4.part will still there forever as leftovers, otherwise.

Instead if yt-dlp works fine, they do not remove anything because there is nothing to be removed.

Video/tormyvancool_YOUTUBE Downloader.lua
tormyvancool: Ok I will.

tormyvancool: Modified as indicated

Video/tormyvancool_YOUTUBE Downloader.lua
cfillion: There's also no way `rm "/project/dir\Videos\*.mp4.part"` does anything beside output `No such file or directory` to stderr... (Unless there's a file literally named `Videos\*.mp4.part` with both the backslash and the star in the filename.)

Video/tormyvancool_YOUTUBE Downloader.lua
tormyvancool: > Line 146 only runs on macOS. It attempts to set the execute bit on yt-dlp_macos in whichever directory is current (unlikely to be yours the first time). So the line is effectively no-op. Only the one at line 145 is effective (ignoring the same injection problem). By the way it's checking `OS == "OSX32"` despite the script (well, the yt-dlp binary) not supporting macOS 32-bit?

Actually even the very first time is my directory.
It creates firs the yt-dlp subtir
then, on it, it will download the yt-dlp.something
Only from that moment on the flag is changed on +x

in can't be otherwise. or it means the script is not executed linearly form line 1 to the last and something si skipped.
But if so, it means something else is not working as it should.
So that "cd " can't be wrong

NOTE: I solved the checking for OSX2 that however it didn't do anything wrong if there.
I have always the same if then else end on a separate file and I copy and paste it.
If the script is installed on any machine: ti works.
And that check is not damaging anything.

Video/tormyvancool_YOUTUBE Downloader.lua
tormyvancool: I'm checking what I can do to get the identical thing I want, by the use of the suggested APIs

and for the "rm" not clue why it worked on my virtual machine then.

Video/tormyvancool_YOUTUBE Downloader.lua
cfillion: > It creates firs the yt-dlp subtir
> then, on it, it will download the yt-dlp.something
>
> in can't be otherwise

That's not in the script's code? `getOS` is called from line 175 and it's the first thing that's done after clearing the console and defining a few variables and functions. Nothing changes the current working directory of REAPER's process.

The `yt-dlp` directory is created and filled when the package is installed by ReaPack. Not when running the script. And ReaPack does so without messing with REAPER's working dir.

Video/tormyvancool_YOUTUBE Downloader.lua
tormyvancool: NOTE in `integer fileindex` it will be nice to understand what's expected. I just used the -1 only

I supposed it should be used in this way  `ResFiles = reaper.EnumerateFiles( ProjDir .. '/Videos/*.mp4.part', 1)`
But unlike the commands I wrote and they are working, this one doesn't :-/

Video/tormyvancool_YOUTUBE Downloader.lua
tormyvancool: > > It creates firs the yt-dlp subtir
> > then, on it, it will download the yt-dlp.something
> > in can't be otherwise
> 
> That's not in the script's code? `getOS` is called from line 175 and it's the first thing that's done after clearing the console and defining a few variables and functions. Nothing changes the current working directory of REAPER's process.
> 
> The `yt-dlp` directory is created and filled when the package is installed by ReaPack. Not when running the script. And ReaPack does so without messing with REAPER's working dir.

yes ReaPack does it because the code is telling to do so, due the lines @provide.
Hence the `yt-dlp` subdir is created, and the script behaves in the way I said.

EDIT: instead if you want the script goes to check if really that subdir is created otherwise alerts and stop running, is something else. But I have not clue (now)how to access the scripts dir to investigate it's existence with Reaper's APIs.

Video/tormyvancool_YOUTUBE Downloader.lua
cfillion: `reaper.EnumerateFiles(ProjDir .. '/Videos', i)` in a loop with i incremented by one each iteration starting from 0 until nil is returned. Check if the returned filename ends with '.mp4.part' and delete if so...

tormyvancool: Ok found the way

```
              a=0
              test = ""
              repeat
              reaper.ShowConsoleMsg(test)
              test = reaper.EnumerateFiles( ProjDir .. "\\Videos\\*.mp4.part", a)
              a = a + 1
              until(test == nil)
```

I will adapt it and implement it

tormyvancool: Sorry I read you after have posted mien one. Ok indeed I'm doing that implementation and some part of the code mus be rewritten although it will be simplified and more efficient

tormyvancool: Committed

Video/tormyvancool_YOUTUBE Downloader.lua
tormyvancool: I need your feedback to work on this.

Video/tormyvancool_YOUTUBE Downloader.lua
tormyvancool: Committed

Video/tormyvancool_YOUTUBE Downloader.lua
tormyvancool: ok no feedback = free interpretation :-)
It means: I found the way to check if that subfolder is created or not.
If not created a message will appear telling that the script MUST be installed via reapack repository.

Once implemented and tested I will commit it.

tormyvancool: Committed

tormyvancool:
> It's an improvement over last time however there are still significant design flaws that I can't accept on a default repository. Most notably the complete absence of any protection against injection which is unchanged from the previous PR.
> 
> In the current state I think it would be better released on your own repository.

I went through all what highlighted and committed the solutions.

Video/tormyvancool_YOUTUBE Downloader.lua
tormyvancool: I reopened this conversation because I see I can perhaps review that function that selects the OSes. I will try to make it better. Give me some time. I will eb back when ready by closing this conversation and update the code

tormyvancool: There is again to work on the `getOS()` function I see. I will work on the path now and I will update the code.

tormyvancool: Ok super. Committed. the path is detected by the Reaper's APIs and it's not more hardencoded

tormyvancool:
Ok I think i've done with this script. All what we discussed and also something more, was done and the "something more" was added. 
So in case something goes wrong, the script shouldn't freeze Reaper indefinitely, and informs the operator stopping itself and let Reaper free of working.

Video/tormyvancool_YOUTUBE Downloader.lua
cfillion: > EDIT: instead if you want the script goes to check if really that subdir

I didn't say that? There is no reason to check whether ReaPack has installed yt-dlp. It either does or the entire package is not installed at all & the user is informed of the error.

I was talking about the broken & redundant chmod which is still not removed.

cfillion:
Going to reject this for now. Thanks for applying some of the smaller suggestions, however the major flaws pointed out in both this and the previous PRs are still not addressed. Shell command execution requires more care than most scripts and must either be done right or not attempted.

tormyvancool:
> Shell command execution requires more care than most scripts and must either be done right or not attempted.

Shell commands are for what it should be done the way I expect. The function you suggested me: doesn't and it creates more issues than benefits.

> I didn't say that? There is no reason to check whether ReaPack has installed yt-dlp. It either does or the entire package is not > installed at all & the user is informed of the error.
>
> I was talking about the broken & redundant chmod which is still not removed.

That's why I asked for the confirmation of what I understood. You came today. I asked 3 days ago. And I added "ok no feedback = free interpretation :-)"  so meanwhile I worked on what I understood.

The redundant `chmod +x` is required since during the installation of the package it seems there is not way to guarantee the `chmod +x `as required to the downloaded binary. And this is was clearly issued during the testing of the previous versions. Indeed yt-dlp was downloaded without that flag (it turned out to me only read mode). Solution: run that command to be sure is +x 

If I ask to the O.S. to "chmod +x" when the file already is in "x" nothing happens. It survives on this and it doesn't require more than some microsecond.

tormyvancool:
NOTE the fact you closed this PR, obliges me to pass again that uncomfortable form.It requires more effort there tan to remove a redundancy on os.execute() with the issues on downloading only the 64 bits etc ...

cfillion:
> Shell commands are for what it should be done the way I expect. The function you suggested me: doesn't and it creates more issues than benefits.

`os.execute` is fine if the input is properly sanitized ([not](https://daviddeley.com/autohotkey/parameters/parameters.htm#WINCRULESDOC) [trivial](https://flatt.tech/research/posts/batbadbut-you-cant-securely-execute-commands-on-windows/) on Windows, not too bad on Unix-likes). `ExecProcess` is easier in this aspect, but of course it comes with limitations (no escape character = worse than `os.execute` for compatibility) and can be tricky to use without understanding it.

> That's why I asked for the confirmation of what I understood. You came today. I asked 3 days ago.

...are you really complaining because I didn't reply within the 2/6 hours between your "I need your feedback to work on this" and "ok no feedback = free interpretation" / "Ok super. Committed." replies?

> The redundant chmod +x is required

Two chmod are redundant. One is enough. Looks like you tried to fix that in #1461 by removing one of them. Good but the one that's left is not going to work without being provided the correct path to the file.

> NOTE the fact you closed this PR, obliges me to pass [...]

You can re-create a PR without going through the form again (same git branch). But please go ahead and release it in your own repository instead. I'm sure the script is useful and mostly works but I can't accept it here (where it's going to be shipped to all users) unless the various packaging and core functionality issues are fully addressed.

- Hardened against command injection
- `chmod` not broken on macOS – since you're testing on Linux: macOS is pretty much the same yet your code is different for some reason
- `yt-dlp` pinned to a specific version (when an update –yours or yt-dlp– breaks something, users need to trust ReaPack's downgrade feature to undo the update without breaking a sweat)
- No multi-version changelog

While at it (optional, out of scope for a packaging review) the code quality could be improved. There's unused leftovers, inconsistencies and typos (such "Downlaod" and "filname")...

tormyvancool:
> > Shell commands are for what it should be done the way I expect. The function you suggested me: doesn't and it creates more issues than benefits.
> 
> `os.execute` is fine if the input is properly sanitized ([not](https://daviddeley.com/autohotkey/parameters/parameters.htm#WINCRULESDOC) [trivial](https://flatt.tech/research/posts/batbadbut-you-cant-securely-execute-commands-on-windows/) on Windows, not too bad on Unix-likes). `ExecProcess` is easier in this aspect, but of course it comes with limitations (no escape character = worse than `os.execute` for compatibility) and can be tricky to use without understanding it.

How to sanitize it more? I have not clue. Or I wasn't make mistakes.
If you notice that os.execute() base don OSes in the beginning are just paths taken by REPAER's APIs.
The same for the variables "Update" and "Video" used into the os.execute() almost at the end of the script.

The input fields (the only oen way I see a potential injection)

- the first one verifies is an URL and it must be valid
- the second one has limited inputs: only alphanumerically characters.

> ...are you really complaining because I didn't reply within the 2/6 hours between your "I need your feedback to work on this" and "ok no feedback = free interpretation" / "Ok super. Committed." replies?

In 3 days (not 6 ours) I was working on it trying also to better understand what you were saying. Do not have feedback I just applied what I did understand. It's not an accusation but it clears out how I was working.
however the control I have pt in place does its job. So ppl will install the script via ReaPack system avoiding to download just the script, copy and paste it and use it, missing the important settings.
 
> Two chmod are redundant. One is enough. Looks like you tried to fix that in #1461 by removing one of them. Good but the one that's left is not going to work without being provided the correct path to the file.

Yes I spotted it after this. That's why I corrected it and opened the PR that you closed now.
 
> You can re-create a PR without going through the form again (same git branch). But please go ahead and release it in your own repository instead. I'm sure the script is useful and mostly works but I can't accept it here (where it's going to be shipped to all users) unless the various packaging and core functionality issues are fully addressed.

That's why I started to ask for HELP everywhere (for the os.execute()). But I don't get the hints I do need to sanitize that damn os.execute()
I'm not born with the knowledge embedded or I wasn't here neither.

Example? tons; For instance how can I reopen a PR by updating the code avodigin that form and avoiding to open a fork?
I cannot edit the code pasting the new one now that's closed.

> * Hardened against command injection

This where I'm still asking for but not answers that really helps me not only to understand but to avoid future mistakes.

> * `chmod` not broken on macOS – since you're testing on Linux: macOS is pretty much the same yet your code is different for some reason

In the new code (3.0) I removed it.

> * `yt-dlp` pinned to a specific version (when an update –yours or yt-dlp– breaks something, users need to trust ReaPack's downgrade feature to undo the update without breaking a sweat)

yt-dlp is not released if it breaks everything. yes can have bugs like Reaper has bugs. But it's not breaking anything.

> * No multi-version changelog

I explained you that the detailed changelog served to keep history on what it was done and ReaPack has not that history since from 1.0 to 30 was published on my repository. The form on RePack site, is a text box that enables to paste the whole history.

> While at it (optional, out of scope for a packaging review) the code quality could be improved. There's unused leftovers, inconsistencies and typos (such "Downlaod" and "filname")...

cfillion:
> How to sanitize it more? I have not clue.

- https://www.gnu.org/software/bash/manual/bash.html#Quoting
- https://daviddeley.com/autohotkey/parameters/parameters.htm#WINCRULESDOC
- https://flatt.tech/research/posts/batbadbut-you-cant-securely-execute-commands-on-windows/

> That's why I corrected it and opened the PR that you closed now.

You removed the chmod that was working and kept the one that didn't.

> yt-dlp is not released if it breaks everything. yes can have bugs like Reaper has bugs. But it's not breaking anything.

When something break, which also includes undesired changes of any kind, users need to be able to easily and reliably downgrade.

https://xkcd.com/1172/

> how can I reopen a PR

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request

> I explained you that the detailed changelog served to keep history

The purpose of changelogs is to inform/warn users of changes between versions when they install or update. There is no purpose for changelogs before the first ReaPack release.

Keeping full development history is git's purpose.

> The form on RePack site, is a text box that enables to paste the whole history.

It says immediately below it: "Changelog for the current version only".

tormyvancool:
> > How to sanitize it more? I have not clue.
> 
> * https://www.gnu.org/software/bash/manual/bash.html#Quoting
> * https://daviddeley.com/autohotkey/parameters/parameters.htm#WINCRULESDOC
> * https://flatt.tech/research/posts/batbadbut-you-cant-securely-execute-commands-on-windows/
> 
> > That's why I corrected it and opened the PR that you closed now.
> 
> You removed the chmod that was working and kept the one that didn't.

oops restored and removed the one inside the os.execute()

However about the quoting: what's need to be quoted? Since practically everything is generated by the REAPER's APIs
Or do you mean to avoid the double -quotes? For instance

from `MainPath = '"' .. CallPath .. dlpLnx .. '"'`
to `MainPath = "'" .. CallPath .. dlpLnx .. "'"` ?

On the pull request: once it's closed I have this
image
I can't edit any longer.

An this
"In the "Branch" menu, choose the branch that contains your commits." is not valid once the PR is closed.

cfillion:
> restored and removed the one inside the os.execute()

👍 (Alternatively could have fixed the os.execute one by giving it the correct path.)

> However about the quoting: what's need to be quoted? 

(Linux&macOS) Single quotes are easier but the possibility of those variables containing single quotes and backslashes themselves has to be handled.

> once it's closed I have this

The grayed out 'edit file' button in the screenshot is a shortcut to a similar button on your fork.

Good luck with your script.

tormyvancool:
> > restored and removed the one inside the os.execute()
> 
> 👍 (Alternatively could have fixed the os.execute one by giving it the correct path.)
> 
> > However about the quoting: what's need to be quoted?
> 
> (Linux&macOS) Single quotes are easier but the possibility of those variables containing single quotes and backslashes themselves has to be handled.
> 
> > once it's closed I have this
> 
> The grayed out 'edit file' button in the screenshot is a shortcut to a similar button on your fork.
> 
> Good luck with your script.

The correct path in the last version was given by the API by Reaper. Indeed I changed the way these variables were generated.

About the quotes: besides the only few quotes I put, all the other parameters (like the path and filenames) are generated by Reaper's APIs. I think these APIs are already managing this issue and sanitizing the content.

PR #1458, opened 2024-11-08
Release Trim left, right or both item edges via mouse and razor v1.5

AZpercussion:
- added option for trimming nearest/farthest edge of item as in the Fade tool script
- added TCP user action
- added font scaling for options window

### DIFF
--- a/Items Editing/az_Trim left, right or both item edges via mouse and razor.lua
+++ b/Items Editing/az_Trim left, right or both item edges via mouse and razor.lua
@@ -1,17 +1,21 @@
 -- @description Trim left, right or both item edges via mouse and razor
 -- @author AZ
--- @version 1.4.1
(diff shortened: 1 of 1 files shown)


Items Editing/az_Trim left, right or both item edges via mouse and razor.lua
```diff
@@ -219,8 +248,17 @@ function OptionsWindow()
       OptDefaults[i] = option
     end -- for
     
+    reaper.ImGui_Text(ctx, '' ) --space
```
cfillion, line 251: `ImGui.NewLine`?

AZpercussion: Ah, thanks for the tip!

PR #1465, opened 2024-11-19
Release Play item or track under mouse until shortcut released v1.0

AZpercussion:


cfillion:
The SWS action produces extra loud sounds or doesn't work properly in heavy projects? Not good, do you have steps to duplicate?

AZpercussion:
> The SWS action produces extra loud sounds or doesn't work properly in heavy projects? Not good, do you have steps to duplicate?

Simple steps:
- Have a massive project with some muted tracks.
- Run action to listen a one of quite items
- Release the key and you will hear a burst of energy of all tracks

You can see the reports here: https://github.com/reaper-oss/sws/issues/766#issuecomment-146724640

I figured out that it depends on Render ahead time + Output latency
Also the SWS action temporarily unmutes all tracks that is freeze up heavy projects, especially if playing is going.

PR #1481, opened 2025-01-11
Release Resample Notes v1.0

jopoet:


cfillion:
A few issues:

- Package name and action filenames should be in sentence case (Resample Notes -> Resample notes) to blend with native REAPER actions and other ReaScripts
- Typo: "**Dulpicate** notes octave up (with fx).lua"
- Are the preset files not meant to be added as actions in the Action List? If they are (eg. manually by the user), then they should have the author prefix ("daodan_").
- The GUI script is added to both the MIDI Editor and Main section of the Action List but the Main script is only added to MIDI Editor. Is this intentional?

jopoet:
@cfillion Thanks for review.

>     * Package name and action filenames should be in sentence case (Resample Notes -> Resample notes) to blend with native REAPER actions and other ReaScripts
Ah, it meant to be like "real" name, "Resample Notes" like in "John Doe" (besides being description). But it's not that important, will fix it. Should I keep "Main" and "GUI" suffixes to keep separation from name? Will "daodan_Resample notes Main.lua" and "daodan_Resample notes GUI.lua" be OK?

>     * Are the preset files not meant to be added as actions in the Action List? If they are (eg. manually by the user), then they should have the author prefix ("daodan_").
They are not loaded so as not to clutter the action list, and yes, if the user wants, he can still load them and use them regular scripts. And since loading is optional and these same files are used as presets in the GUI, I decided not to add a prefix, because I thought that it would take up space in vain, cutting off the real name of the preset. But now I tried and it turned out not to be such a big problem. I'll rename it.

>     * The GUI script is added to both the MIDI Editor and Main section of the Action List but the Main script is only added to MIDI Editor. Is this intentional?
Yes. Reasons:
- In case user wants to add button for GUI script to any toolbar instead of MIDI Editor's one
- User forget or don't know that script uses MIDI Editor or forget to switch to MIDI Editor section in Action list. He still can type "resample notes" in main action list and run GUI. GUI script doesn't care if MIDI Editor is opened or not but if user click on "RUN" button when no MIDI Editor opened he will get message (from main script) that there should be selected notes from MIDI editor.
- In future updates (maybe) there will be cases when opened MIDI Editor is not necessary for example such as set to use time selection and ignore note selection or inline midi editor used.

cfillion:
> Ah, it meant to be like "real" name, "Resample Notes" like in "John Doe" (besides being description). But it's not that important, will fix it.

"Resample notes" being verb + noun, it reads more like a generic action to perform (same as eg. "save project") rather than the intended proper noun (which would be OK to capitalize)...

> Should I keep "Main" and "GUI" suffixes to keep separation from name? Will "daodan_Resample notes Main.lua" and "daodan_Resample notes GUI.lua" be OK?

Perhaps "daodan_Resample notes.lua" and "daodan_Resample notes (GUI).lua" or "daodan_Resample notes - GUI.lua"? Implying that the primary one to use is the former and GUI is a variant.

> Yes. Reasons:

OK!

jopoet:
@cfillion Done!
Can you please merge all commits to one named something like "correct file names, update files" so it doesn't look so messy?

jopoet:
Oops, checks failed. It seems I did something wrong. Can you please help?
upd. Yap, replaced the code, but forgot about noindex. Returned it back.
upd2. Ok, I don't understand what's going on. 
"1) MIDI Editor/daodan_Resample notes.lua failed:
  file not found 'daodan_Resample notes/daodan_Resample notes.lua'"
But file is here, right?

cfillion:
`daodan_Resample Notes/daodan_Resample notes.lua` != `daodan_Resample notes/daodan_Resample notes.lua`

I also fixed `goback` for non-Windows: Windows tolerates forward slashes for separating directories, but Linux and macOS allow backward slashes within filenames.

jopoet:
> `daodan_Resample Notes/daodan_Resample notes.lua` != `daodan_Resample notes/daodan_Resample notes.lua`
> 
> I also fixed `goback` for non-Windows: Windows tolerates forward slashes for separating directories, but Linux and macOS allow backward slashes within filenames.
@cfillion 
Good, thank you so much. But I see that you also changed "Resample notes Presets" to "Presets". Why? That's no good because now GUI script won't find those presets until edited. And besides, I specifically chose this name in case the user puts the files directly into the \Scripts\ directory (when installing not via repack), without creating an additional subdirectory. "Resample notes Presets" immediately makes it clear what these files refer to, when just "Presets" folder inside \Scripts\ can be anything, and besides, it can be used for something else (another script). Can you please change it back?

cfillion:
> Can you please change it back?

Certainly. I renamed it because "daodan_Resample notes/Resample notes Presets/daodan_Resample notes XYZ.lua" had so many "Resample notes".

Will push a v1.0.0-1 to undo that. EDIT: #1484.

PR #1485, opened 2025-01-17
Release Update region start to playhead position by ID v1.2

edsonsantoro:
v1.2 Update region start to playhead position by ID

### DIFF
--- a/Regions/edsondelsantoro_Update region start to playhead position by ID.lua
+++ b/Regions/edsondelsantoro_Update region start to playhead position by ID.lua
@@ -1,0 +1,80 @@
+-- @description Update region start to playhead position by ID
+-- @author Edson Del Santoro
+-- @version 1.2
+-- @changelog v1.2 Update region start to playhead position by ID
+-- @about
+--   # Update region start to playhead position by ID
+--
+--   --   # Update region start to playhead position by ID
+--   --   This script updates the start time of a region specified by its ID (as shown in the Region/Marker Manager). 
+--   --   It keeps the region's end time intact and adjusts only the start.
+--
+--   1. Set the playhead to the start position that you want.
+--   2. Execute this script.
+--   3. Enter the Region ID that you want to update Start timming.
+--
+--   That's it.
+
+
+SCRIPT_NAME = "Update region start to playhead position by ID"
+
+reaper.ClearConsole()
+
+function Msg(variable)
+  reaper.ShowConsoleMsg(tostring(variable) .. "\n")
+end
+
+-- Get playhead position
+playhead_position = reaper.GetCursorPosition()
+
+-- Get total number of markers and regions
+retval, num_markers, num_regions = reaper.CountProjectMarkers(0)
+
+if num_regions == 0 then
+  reaper.ShowMessageBox("No regions found in the project.", SCRIPT_NAME, 0)
+  return
+end
+
+-- Ask the user for the ID of the region to update
+retval, user_input = reaper.GetUserInputs(SCRIPT_NAME, 1, "Enter Region ID (as shown in Region/Marker Manager):", "")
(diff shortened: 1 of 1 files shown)


Regions/edsondelsantoro_Update region start to playhead position by ID.lua
```diff
@@ -0,0 +1,80 @@
+-- @description Update region start to playhead position by ID
```
MichaelPilyavskiy, line 1: Bad naming. Better: "Set region start to edit cursor, define region by ID"

Regions/edsondelsantoro_Update region start to playhead position by ID.lua
```diff
@@ -0,0 +1,80 @@
...
+-- @author Edson Del Santoro
+-- @version 1.2
+-- @changelog v1.2 Update region start to playhead position by ID
```
MichaelPilyavskiy, line 4:
```
-- @changelog
--   + initial release
```

Regions/edsondelsantoro_Update region start to playhead position by ID.lua
```diff
@@ -0,0 +1,80 @@
...
+region_marker_index = selected_region.region_marker_index
+rgn_name = selected_region.rgn_name
+
+-- Begin undo-block
+reaper.Undo_BeginBlock2(0)
+
+reaper.SetProjectMarkerByIndex(0, region_marker_index, true, new_start, rgn_end, selected_region.rgn_index, rgn_name, 0)
+
+-- End undo-block
+reaper.Undo_EndBlock2(0, SCRIPT_NAME, -1)
```
MichaelPilyavskiy, line 77: Use 0xFFFFFFFF instead -1

https://www.reaper.fm/sdk/plugin/reaper_plugin.h


```
UNDO_STATE_ALL 
UNDO_STATE_ALL 0xFFFFFFFF 
UNDO_STATE_TRACKCFG 1 // has track/master vol/pan/routing, ALL envelopes (matser included) 
UNDO_STATE_FX 2  // track/master fx 
UNDO_STATE_ITEMS 4  // track items 
UNDO_STATE_MISCCFG 8 // loop selection, markers, regions, extensions! 
```

edsonsantoro (commented):
Thanks @MichaelPilyavskiy for these improvements.

Regions/edsondelsantoro_Update region start to playhead position by ID.lua
Buy-One: Does the choice between -1 and 0xFFFFFFFF affect the result or reliability?
-1  [was recommended by Schwa](https://forum.cockos.com/showpost.php?p=2090533&postcount=27) and as far as i understand in C that's one and the same

MichaelPilyavskiy: > Does the choice between -1 and 0xFFFFFFFF affect the result or reliability? -1 [was recommended by Schwa](https://forum.cockos.com/showpost.php?p=2090533&postcount=27) and as far as i understand in C that's one and the same

It is not the same. -1 stay for "all byte set" while 0xFFFFFFFF stay for "first 8 bytes set", I remember I got different results somewhere using -1, nothing serious but I don`t remember where it differs I just remember 0xFFFFFFFF worked as expected and -1 did not. I use it everythere since that time, maybe they changed that in API in the past (3+ years ago) and that was really rarely undocumented corner case.

cfillion: 4 bytes*

In this case it's equivalent: `Undo_EndBlock2`'s `extraflags` is a 32-bit integer so -1 and 0xFFFFFFFF both set all 32 bits.

PR #1486, opened 2025-01-17
Release Set render matrix by comparing track names and region names with subfolder v1.0

edsonsantoro:


MichaelPilyavskiy:
Sorry, but I don`t see any difference with Mordi`s script except script name, indents and couple of comment and header tweaks

PR #1489, opened 2025-01-19
Release Floating Solo Toolbar v1.0

Neftovsky:


cfillion:
- Scripts should use sentence case to blend well with native actions and other scripts: "Floating solo toolbar", perhaps "Neft's floating solo toolbar" (but there's already neftofsky_ shown in front, so it'd be a bit redundant?), etc...
- Typo: "-aves" in the `@about` text
- Public ReaImGui-based scripts must enable backward compatibility by specifying which API version they are written for (see examples at https://forum.cockos.com/showthread.php?t=250419)

Suggestion: it might be a good idea to save the settings in the same folder as the script's file rather than hard-coded to the resource path's root.

cfillion:
To enable backward compatibility, it's:

```lua
package.path = reaper.ImGui_GetBuiltinPath() .. '/?.lua'
require 'imgui' '0.X.Y'
```

Since the script uses the now-removed DestroyContext function, it looks like it was written for an older version of ReaImGui? Specify which one in the `require` line so that it can work with today's (0.9) and tomorrow's versions.

Tracks/Neftovsky_Floating Solo Toolbar.lua
```diff
@@ -0,0 +1,241 @@
...
+--   - Links each button to a specific track in the project
+--   - Toggles the solo state for a track in the project when a button is clicked
+--   - Allows changing the names of tracks and buttons by right-clicking on the buttons
+--   - Enables adding new buttons to the toolbar
+--   - Saves settings, including button labels and track names
+-- @requires
+--   ReaImGui: API version 0.8 or later
+-- @requires
+--   REAPER 6.0 or later
+
```
MichaelPilyavskiy, line 13: It is better to use something like:

```
app_vrs = tonumber(reaper.GetAppVersion():match('[%d%.]+'))
check_vrs = 6.0
if app_vrs < check_vrs then return reaper.MB('This script require REAPER '..check_vrs..'+','',0) end
local ImGui

if not reaper.ImGui_GetBuiltinPath then return reaper.MB('This script require ReaImGui extension','',0) end
package.path =   reaper.ImGui_GetBuiltinPath() .. '/?.lua'
ImGui = require 'imgui' '0.9.3.2'
```

Tracks/Neftovsky_Floating Solo Toolbar.lua
```diff
@@ -0,0 +1,241 @@
...
+            end
+            RenderMenuPopup()
+            reaper.ImGui_End(imgui)
+        end
+
+        if not open then
+            window_open = false
+        end
+    else
+        reaper.ImGui_DestroyContext(imgui)
```
MichaelPilyavskiy: DestroyContext is not required for 0.9.3.2

Tracks/Neftovsky_Floating Solo Toolbar.lua
```diff
@@ -0,0 +1,241 @@
...
+        local settings = StringToTable(content)
+        if type(settings) == "table" and settings.buttons and settings.order then
+            buttons = settings.buttons
+            button_order = settings.order
+        end
+    end
+end
+
+local function SaveSettings()
+    local file = io.open(save_file_path, "w")
```
MichaelPilyavskiy: This is really bad idea. At the you have NeftToolbarSettings.lua, which kinda  "wtf is that", because it is not a script, while it is marked as a script. I would recommend to use Get(Set)ExtState() for things like that. It is better to use base64 encoding/decoding if you store multiline strings (very close to your case because you store serialized table).

Neftovsky:
I'm sorry for this mess. This is the first time I've posted something here. And anyway, this is my first script. I'm writing it with GPT. See the latest fix

cfillion:
The action filename still wasn't in sentence case: fixed via #1495.

PR #1497, opened 2025-02-03
Release Apply track volume to Items v1.0

Neftovsky:


cfillion:
"Apply track fader volume to all items in selected tracks and reset faders to 0dB" already exists (by amagalma).

Also "Apply track volume to items" is incorrect: this new one sets the active (only!) take volume instead of item volume.

Neftovsky:
> "Apply track fader volume to all items in selected tracks and reset faders to 0dB" already exists (by amagalma).
> 
> Also "Apply track volume to items" is incorrect: this new one sets the active (only!) take volume instead of item volume.

The amagalma's script changes the volume of the pre-fx items and this is displayed on the items themselves. My script changes the volume in the properties of the items without affecting the pre-fx fader. Before that. It looks more calm visually. And you can also change the pre-fx after

cfillion:
It changes the active take's volume. Not the item's volume. Both take and item volumes are pre-FX.

(Wouldn't it be less surprising to the user if it changed all takes?)

Neftovsky:
Thanks for the comments! I've updated the script: now it changes the volume of all takes in the items, not just the active one

PR #1505, opened 2025-02-27
Release MaCCLane v0.1.0 (Initial Release)

BenTalagan:
Very first version of MaCCLane, will probably enter first refinement loop with forum users : 

https://forum.cockos.com/showthread.php?t=298707


BenTalagan:
Thank you for the super fast merge !

PR #1518, opened 2025-03-16
Update MaCCLane v0.1.6 > v0.1.7

BenTalagan:
Linux Fix for MacLLane, changelog :

- [Bug Fix] [Linux] Wrong coordinate system leading to widget not appearing at all

Please note that I'm committing a local .gitignore for my own talagan_MaCCLane sub-directory (with local effect, thus), because I'm experimenting a new workflow to ease the interaction between git / local dev branches / REAPER script directories, where I can develop directly in my local git. I hope I'm not too intrusive adding such a file or violating the cleanliness rules for the ReaTeam repositories. In fact I was a bit surprised there wasn't any .gitignore file at all for the full repo, but considering that plugins are mainly added by the web tool it also feels a bit logical.


cfillion:
Git supports local ignore rules via `.git/info/exclude` instead of a public `.gitignore`.

BenTalagan:
Awesome, thanks Christian. I wasn't aware of this feature. I've removed the added .gitignore file, and you can merge the fix.

PR #1528, opened 2025-03-25
Update MaCCLane v0.1.7 > v0.2.0

BenTalagan:
Update for MaCCLane. Changelog : 

 - [Major Feature] Current tab / tab state
  - [Major Feature] Record modes for every module
  - [Feature] Added "Reveal template folder" in (+) menu
  - [Feature] Added Return / Escape key handling when editing tab name
  - [Feature] Added "Return key behaviour when editing tab name" customization
  - [Feature] Added "New Full Recording Tab" entry to menu
  - [Feature] Added debug tools
  - [Bug Fix] Setting ME grid does not work if global grid setting is set to "Measure"
  - [Bug Fix] MaCCLane crash if tab editor grid module is set to custom and closing midi editor (thanks @Seventh Sam)


MIDI Editor/talagan_MaCCLane/lib/inspect.lua
```diff
@@ -0,0 +1,358 @@
...
+
+local shortControlCharEscapes = {
+   ["\a"] = "\\a", ["\b"] = "\\b", ["\f"] = "\\f", ["\n"] = "\\n",
+   ["\r"] = "\\r", ["\t"] = "\\t", ["\v"] = "\\v", ["\127"] = "\\127",
+}
+local longControlCharEscapes = { ["\127"] = "\127" }
+for i = 0, 31 do
+   local ch = char(i)
+   if not shortControlCharEscapes[ch] then
+      shortControlCharEscapes[ch] = "\\" .. i
```
cfillion, line 73: Lua has a built-in feature to format any value as a valid Lua literal: `string.format('%q', any_value)`.

EDIT: "Any" being just booleans, nil, numbers, and strings.

BenTalagan: Thanks for the info, Christian ! To be frank, since it is an external library, I did not took the time to dig inside its implementation. This was directly taken from : https://github.com/kikito/inspect.lua which is a bit old and might be worth a few updates.

PR #1539, opened 2025-04-10
Update cfillion_Lua profiler.lua

nickvonkaenel:
formatNumber returns a string and string.format is expecting a number. Lua handles this fine when the number is below 1000 but when formatNumber adds a comma to the number string for numbers over 1,000 it causes an error. Changing %d to %s fixes this.


cfillion:
Thanks! Applied as https://github.com/cfillion/reascripts/commit/c2f5a64cce6542f4080ce3211e013805c4fb43cd in my development repository. (ReaTeam repos are for releases only.)

EDIT: Released as v1.1.5.

PR #1541, opened 2025-04-11
Release Set pre-roll duration to 1 bar v1.0

JMWalkman:


cfillion:
Action variations like this must be packaged together instead of as separate packages.

This also allows for having a single source file that reads its own installed filename to determine which value to set: https://github.com/cfillion/reapack-index/wiki/Examples#multiple-slots.

Using the GUI upload tool:

1. Set the main package file's resource type to "Don't install"
2. Click on "Add additional files"
3. Set that file's install filename as desired
4. Set that file's file source to be the main package file
5. Repeat from step 2 for each additional variation

PR #1545, opened 2025-04-11
Release Set pre-roll duration to double the current value v1.0

JMWalkman:


cfillion:
This one could probably be added to the same package along with the other "Set pre-roll duration to X bars" actions.

JMWalkman:
> This one could probably be added to the same package along with the other "Set pre-roll duration to X bars" actions.

Thank you, I'll try that

PR #1553, opened 2025-04-27
Release Mouse Buddy v3.0.2

FunkybotsEvilTwin:


Cursor/funkybot_Mouse Buddy.lua
```diff
@@ -0,0 +1,805 @@
...
+  "Double Click",
+  "Right Drag",
+  "Middle Click",
+  "Middle Drag"
+}
+
+-----------------------------------------------------------
+-- Create an ImGui context with docking enabled.
+-----------------------------------------------------------
+local ctx = r.ImGui_CreateContext("Mouse Buddy 3.0", r.ImGui_ConfigFlags_DockingEnable())
```
cfillion: 1. Publicly released ReaImGui scripts must enable backward compatibility by specifying which version they are written for. See snippets at https://forum.cockos.com/showthread.php?t=250419.
  The ImGui.* (instead of r.ImGui_) convention is optional but recommended for enhanced backward compat. More details & script to auto-convert existing code at https://forum.cockos.com/showpost.php?p=2775338.
3. `ConfigFlags_DockingEnable` has been enabled by default for a few years. Explicitly specifying it here overrides the user's corresponding setting. (If that is the intention, then that is fine.)
4. The context name is the key for window state storage. I wouldn't recommend adding the version into it because changing it will reset that storage when it changes.

Cursor/funkybot_Mouse Buddy.lua
```diff
@@ -0,0 +1,805 @@
...
+if r.ImGui_GetIO then
+  local io = r.ImGui_GetIO(ctx)
+  if io then io.ConfigFlags = io.ConfigFlags | r.ImGui_ConfigFlags_DockingEnable() end
+end
```
cfillion: There has never been a `ImGui_GetIO` in ReaImGui. Even if it did, there's no way it could return a table. And even if it did, assigning to it without calling some king of Set function wouldn't do anything...

Is this AI-generated?

The correct way to do what these four lines appear to be doing is:
```lua
ImGui.SetConfigVar(ctx, ImGui.GetConfigFlags(ctx) | ImGui.ConfigFlags_DockingEnable)
```
...which is equivalent to what `CreateContext` already does immediately right above.

Cursor/funkybot_Mouse Buddy.lua
```diff
@@ -0,0 +1,805 @@
+-- @description Mouse Buddy
```
cfillion, line 1: To match the convention of other scripts and native REAPER actions:
```suggestion
-- @description Mouse buddy
```

FunkybotsEvilTwin:
Hi @cfillion, please review the changes and let me know if good. I'm a relative novice here. 

And to answer an earlier question: yes, ChatGPT helped with portions of the script. There's some hand-rolled code from ToursitKiller that is the basis of this script and I built upon with ChatGPT to help with scripting additional features on top of that and ReaImGui code. This is noted in the comments.

cfillion:
The indentation of the metadata header got messed up and there are 30 `r.ImGui_*` remaining (ignoring the first two which are OK)...

Blindly AI-generated scripts and code not submitted by the original author are generally [not allowed](https://github.com/ReaTeam/ReaScripts/wiki/Guidelines) here. Did Touristkiller give permission or a license for their part of the code to be used this way?

FunkybotsEvilTwin:
Tourist Killer gave the ok in the Mouse Buddy thread in the forum, post
here:

Cockos Incorporated Forums - View Single Post - Script: "Mouse Buddy" Mouse
Modifier Context Hints - Thanks TK! Now at Version 3!
<https://forum.cockos.com/showpost.php?p=2843391&postcount=19>

I'm totally fine not having the script on ReaPack. I only even tried
because multiple users in that thread kept requesting it. I'm just as happy
to say "sorry, I tried but ReaScript doesn't allow ChatGPT generated
ReaScripts".

cfillion:
To clarify, it's only when the AI's often subtly incorrect output is blindly trusted that it becomes a problem (just testing the script "it seems to work" is not enough: human reasoning on the code is required).

With proper understanding of the code and the AI artifacts cleaned up, a good AI-assisted ReaScript can be indistinguishable quality-wise from a 100% human one. Those are fine. Keyword being AI-assisted, not AI-generated.

(Current AI tech is unable to reason. It only predicts the most likely "next word" over and over.)

You'd be welcome to *manually* fix the AI's blunders and submit the correction.

`get_mouse_context` for example contains some pretty suspicious code... (`BR_GetMouseCursorContext` always returns lowercase strings, yet the code checks in case it might not be strings then converts to uppercase with `string.upper`, only to re-convert to lowercase immediately after but this time exhaustively using hard-coded values... except in the second `native_ctx` branch where it uses `string.lower` to do the same thing. Except for `segment = "TRACK" which stays upper-cased unless window is "tcp".)

FunkybotsEvilTwin:
Thanks for the input.

I can read the scripts insofar as being able to follow along with the comments and logic, but the coding/syntax is lost on me beyond "does it work".

Appreciate your guidance!

Regarding the upper-case string thing...I vaguely recall that being intentional to help make the matching easier to the .HTML headers. If I'm not remember wrong (this was months ago now), I think we (me and the machine) were getting errors trying match lower-case (from the API) to sentence-case (in the .HTML file), so some level of conversion was needed. 

But I'm just a monkey at a typewriter with this stuff. Then again, I've gotten worse results with human developers at my dayjob! ;)

Anyway, thanks again!

PR #1556, opened 2025-05-01
Release Peak Envelope Generator v1.0

saul-l:


Envelopes/saull_Peak Envelope Generator.lua
```diff
@@ -0,0 +1,277 @@
+-- @description Peak Envelope Generator
```
cfillion, line 1:
```suggestion
-- @description Peak envelope generator
```
...to match the convention of native REAPER actions and other scripts. (Same with the filename.)

Envelopes/saull_Peak Envelope Generator.lua
```diff
@@ -0,0 +1,277 @@
...
+
+
+if not reaper.ImGui_GetBuiltinPath then
+  reaper.MB("ReaImGui required. Please install it.", "Install ReaImGui", 0)
+  if not pcall(function() reaper.ReaPack_BrowsePackages( "ReaImGui: ReaScript binding for Dear ImGui" ) end) then
+    reaper.MB("Attempting to open ReaPack failed. Visit https://reapack.com/", "No ReaPack found", 0)
+    if reaper.GetOS() == "Win32" or reaper.GetOS() == "Win64" then
+      os.execute('start https://reapack.com')
+    else
+      os.execute('open https://reapack.com')
```
cfillion, line 27: On Linux it's `xdg-open`.

(There's `CF_ShellExecute('url')` in SWS that's cross-platform & avoids using the system shell but I don't think it would be appropriate here since it'd be yet another dependency that could be missing.)

(`ReaPack_BrowsePackages` itself won't emit any errors, but a simpler presence check could be `if not ReaPack_BrowsePackages then`.)

Envelopes/saull_Peak Envelope Generator.lua
```diff
@@ -0,0 +1,277 @@
...
+  
+ -- INIT
+ function Apply()
+    countSelectedItems = reaper.CountSelectedMediaItems(0)
+    reaper.PreventUIRefresh(1) 
+    reaper.Undo_BeginBlock()
+    if smooth then reaper.Main_OnCommand(40331,0) end
+    selectedItems = SaveSelectedItems()
+    CalculateEnvelope()
+    reaper.Undo_EndBlock("peakEnvelope", -1)
```
cfillion, line 147: This string is the undo point name shown to the user; wouldn't it be better if it matched the script's name "Peak envelope generator"?

```suggestion
    reaper.Undo_EndBlock("Peak envelope generator", -1)
```

PR #1561, opened 2025-05-03
Release Smart Tempo Ripple v1.5

muorsic:
Version 1.5 April 29th 2025
-- Prevent UI refresh

### DIFF
--- a/Various/muorsic_Smart Tempo Ripple.lua
+++ b/Various/muorsic_Smart Tempo Ripple.lua
@@ -1,0 +1,465 @@
+-- @description Smart Tempo Ripple
+-- @author muorsic
+-- @version 1.5
+-- @changelog
+--   Version 1.5 April 29th 2025
+--   -- Prevent UI refresh
+-- @about
+--   Features
+--   Captures original values and sets timebase to "time" before opening Reaper’s dialog (so no beats are added/subtracted when changing signatures).
+--   Compares new values to old ones.
+--   Handles tempo changes before time signatures.
+--   Fixes partial measures by inserting the correct time signature.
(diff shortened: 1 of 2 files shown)


cfillion:
The script's name should be in sentence case to match native REAPER actions and other ReaScripts. However it's unclear what is the actual desired name of the script:

- "Smart tempo ripple" as per the current filename/description
- "Smart tempo changes" as per the undo point name (`Undo_EndBlock`)
- "Smart tempo and time signature changes" as per the forum thread title
- "Smart tempo changes – Ripple time signatures" as per the forum thread body

muorsic:
Hi cfillion,
thanks for reviewing this. yes it does use:
Tempo envelope: Insert tempo/time signature change marker at edit cursor...
But we can't use / or : in the name . And is this too long?
Tempo envelope- Insert tempo time signature change marker at edit cursor, avoiding partial measures and added beats.
Also - will I update the script here or on the ReaPack website?

cfillion:
I meant that it should use the same capitalization style as the existing actions (REAPER uses sentence case rather than Title Case), not that it should match "Tempo envelope: Insert tempo/time signature change marker at edit cursor" specifically.

(But if that's the name you prefer, it's perfectly fine even if it's on the longer side.)

Any name you choose should also be consistent between the filename, `@description`, undo point, and preferably how it's announced on the forum. So that it's easy to find.

If a user reads "Smart tempo changes – Ripple time signatures" on the forum and then tries to search that in ReaPack, they won't find it if it's named "Smart tempo ripple" here.

You can edit the files directly here *or* restarting on reapack.com/upload. Whichever you're more comfortable with.

muorsic:
Hi cfillion,
how does it look now?

Thanks and sorry for the trouble.
btw, thanks for all your amazing scripts !

PR #1567, opened 2025-05-10
Release Monitor On when Midi Instrument Loaded v1.0

grayson-solis:


Video/graysonsolis_Monitor on when midi instrument loaded.lua
```diff
@@ -0,0 +1,55 @@
...
+
+function check_monitor()
+  for i = 0, reaper.CountTracks(0) - 1 do
+    local tr = reaper.GetTrack(0, i)
+
+    -- Check if the track has an instrument plugin
+    if reaper.TrackFX_GetInstrument(tr) ~= -1 then
+
+      -- If monitoring is not enabled, enable it
+      if reaper.GetMediaTrackInfo_Value(tr, "I_RECMON") ~= 1 then
```
cfillion, line 41: This overrides record monitoring in tape mode (I_RECMON = 2). Is that intentional?

grayson-solis: Overrides everything. Figured people would want this turned on when working with a lot of vst instruments like I do. Whenever I want access to this setting again I just remove the vst from the track and i can then select record monitoring in tape mode. Or create a new track to use this function.

PR #1574, opened 2025-05-10
Release Unselect Topmost Track in Selection v1.0

grayson-solis:


cfillion:
"Topmost" is more frequently used when referring to Z-Order. Perhaps "Unselect first selected track" would be a better name easier to find?

There is already:
- "Unselect first/last selected item" by FnA
- "Xenakios/SWS: Select first of selected tracks" in SWS

"Unselect first/last selected track" could be useful too (two actions in the same package).

PR #1590, opened 2025-06-02
Release jd_Kanban v1.0.0

JulyDay0788:


cfillion:
Some suggestions:

`GetProjectPath` refers to the project's Media directory which can be global depending on user settings instead of per-project as one might expect.

It might be useful to store the tasks in the project file itself instead of an adjacent file (`SetProjExtState`).

Also it seems the script doesn't gracefully handle the active project not being saved yet, or changing while the script is open (tabs, Open...).

Buy-One:
In my instance of the script i've modded the line `local project_path = reaper.GetProjectPath()` into

```
local proj, project_path = reaper.EnumProjects(-1)
	if #project_path == 0 then
		if reaper.MB('      The project must be saved\n\n      before the script can work.\n\n\t     Save now?', 'PROMPT', 4) == 6 
		then
		reaper.Main_SaveProject(0, false) -- forceSaveAsIn false, but for unsaved project doesn't matter
		proj, project_path = reaper.EnumProjects(-1)
			if #project_path == 0 then -- user closed Save dialogue without saving
			return reaper.defer(function() do return end end) 
			end
		else return -- user declined the prompt
		reaper.defer(function() do return end end)
		end
	end
```

PR #1593, opened 2025-06-04
Release Kanban v1.0.1

JulyDay0788:
### Added
- Project change detection and automatic task reloading when switching between projects
- Graceful handling of unsaved projects with user prompt to save before proceeding
- Support for storing tasks directly in project files using REAPER's extended state system
- Automatic project dirty marking when tasks are modified

### Changed
- **BREAKING**: Replaced INI file storage with REAPER's `SetProjExtState`/`GetProjExtState` system
- Tasks are now stored inside the project file instead of adjacent INI files
- Improved project path handling using `EnumProjects(-1)` instead of `GetProjectPath()`
- Enhanced project state monitoring with `GetProjectStateChangeCount()`

### Fixed
- Resolved issues with `GetProjectPath()` referring to global media directory instead of per-project paths
- Fixed script behavior when active project changes while script is running
- Improved handling of project tab switching and "Open..." dialog interactions

### DIFF
--- a/Project Properties/jd_Kanban.lua
+++ b/Project Properties/jd_Kanban.lua
@@ -1,6 +1,23 @@
 -- @description Kanban
 -- @author Julyday
--- @version 1.0.0
+-- @version 1.0.1
+-- @changelog
+--   ### Added
+--   - Project change detection and automatic task reloading when switching between projects
+--   - Graceful handling of unsaved projects with user prompt to save before proceeding
+--   - Support for storing tasks directly in project files using REAPER's extended state system
+--   - Automatic project dirty marking when tasks are modified
+--
+--   ### Changed
+--   - **BREAKING**: Replaced INI file storage with REAPER's `SetProjExtState`/`GetProjExtState` system
+--   - Tasks are now stored inside the project file instead of adjacent INI files
+--   - Improved project path handling using `EnumProjects(-1)` instead of `GetProjectPath()`
+--   - Enhanced project state monitoring with `GetProjectStateChangeCount()`
+--
+--   ### Fixed
+--   - Resolved issues with `GetProjectPath()` referring to global media directory instead of per-project paths
+--   - Fixed script behavior when active project changes while script is running
+--   - Improved handling of project tab switching and "Open..." dialog interactions
 -- @about
 --   # Kanban
 --   Graphical interface for managing project tasks in REAPER as a Kanban board.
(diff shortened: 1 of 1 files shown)


Buy-One:
Scroll doesn't seem to work

JulyDay0788:
> Scroll doesn't seem to work

It works for me. The scroll bar, as well as the ability to scroll, appears when enough cards have accumulated.

cfillion:
It only works at the original requested `window_h` not at smaller sizes. And when it does work the text and close button of the first visible item doesn't move until completely offscreen.

EDIT: Also here `delta` of one mousewheel tick is 120/-120 so 1 tick scrolls way more than expected (by +3600px/-2966px) (meanwhile dragging the scrollbar goes slower than the mouse movement).

https://github.com/user-attachments/assets/04b26136-c6b5-42ca-b3cb-73fa6c1129c8

Also unclear why it queries things like `GetProjectStateChangeCount` without ever using that, or refusing to work without having saved the project. 

> Enhanced project state monitoring with GetProjectStateChangeCount()

Is this AI generated?

Buy-One:
> refusing to work without having saved the project

That was relevant for the old method of storing data, i.e. in the external file, because the script used or was supposed to use project path which is non-existent before project is saved

> The scroll bar, as well as the ability to scroll, appears when enough cards have accumulated.

My max screen resolution is 1366 x 768 and this is how it looks on my end. The UI doesn't allow scrolling past the last task to add a new one in the longest column

image

cfillion:
@Buy-One Yup: the script assume the OS will allow the full 1280x900 window size that it requests and that the user never resizes it.

> That was relevant for the old method of storing data

Sure, yet this new version removed that old method while adding adding the check... along with a dedicated line in the changelog so it doesn't sound like just a forgotten leftover...

> Graceful handling of unsaved projects with user prompt to save before proceeding

Buy-One:
@JulyDay0788 I've submitted a pull request in your [repo](https://github.com/JulyDay0788/ReaScripts/pull/1) with the modified script. Please take a look.

JulyDay0788:
@Buy-One 
Thank you for helping to improve the script! I would like to be able to create tasks without a deadline. Now, when I create a new task, the current time is set there by default and the script does not allow me to create tasks with the current time. Thus, it takes more mouse clicks than I would like to create a task.

Buy-One:
>I would like to be able to create tasks without a deadline

You can, i described how in the block comment below the header

>if date field is empty time is ignored

Which creates a task without a deadline. So to have a task without a time stamp it suffices to delete the date.
Plus there's a comment in the top bar of the task dialogue window `[date and time are optional]`

JulyDay0788:
@Buy-One Then it would be good to add the logic "If the user has specified the past or present tense, create a task without a deadline." Or leave the date field empty by default. So that there is no need to additionally press buttons while making task lists. 
And it would be good to remove the "ReaScript console output" window, because this also creates the need to perform additional actions to close this window after closing the window with Kanban.

Buy-One:
> Or leave the date field empty by default
> remove the "ReaScript console output" window

Done. However it's still not the public version of the code, although i didn't mean to leave the ReaScript console, just missed one line.

PR #1606, opened 2025-07-04
Release Auto Change Video v1.0

grayson-solis:


Video/graysonsolis_Auto change video.lua
```diff
@@ -0,0 +1,124 @@
...
+local PreventUIRefresh      = reaper.PreventUIRefresh
+local defer                 = reaper.defer
+local time_precise          = reaper.time_precise
+local GetSelectedTrack      = reaper.GetSelectedTrack
+local CountSelectedTracks   = reaper.CountSelectedTracks
+
+local function isVisible(tr)
+  while tr do
+    tr = GetParentTrack(tr)
+    if tr and GetMediaTrackInfo(tr, "I_FOLDERCOMPACT") == 1 then
```
cfillion, line 34: This only checks for the first collapsed level (I_FOLDERCOMPACT==1) but ignores the second "fully collapsed" level (I_FOLDERCOMPACT==2) – is that the correct behavior?

PR #1607, opened 2025-07-04
Release List FX Parameter Indexes v1.0

grayson-solis:


Development/graysonsolis_List FX Parameter Indexes.lua
```diff
@@ -0,0 +1,11 @@
+-- @description List FX Parameter Indexes
+-- @author Grayson Solis
+-- @version 1.0
+-- @metapackage
```
cfillion: Since the script is made of a single file, this does not need to be a metapackage (separate packaging metadata file & script file).

PR #1570, opened 2025-05-10
Release Toggle Group v1.0

grayson-solis:

PR #1610, opened 2025-07-04
Release Toggle Group v1.0

grayson-solis:


Items Properties/graysonsolis_Toggle Group.lua
```diff
@@ -0,0 +1,53 @@
...
+        reaper.SetMediaItemInfo_Value(reaper.GetSelectedMediaItem(0, i), "I_GROUPID", group_id)
+    end
+
+elseif has_grouped and not has_ungrouped then
+    for i = 0, sel_count - 1 do
+        reaper.SetMediaItemInfo_Value(reaper.GetSelectedMediaItem(0, i), "I_GROUPID", 0)
+    end
+
+else
+    local new_id = reaper.GetProjectStateChangeCount(0) + 1
```
cfillion: `GetProjectStateChangeCount` ([still](https://github.com/ReaTeam/ReaScripts/pull/1570#discussion_r2083302559)) does NOT return a reliable next available group ID.

grayson-solis: - Fixed group ID thing
- Made sure if there is a single item selected that is ungrouped, the script wont work on it (doesnt make sense to add one thing to a group; gets confusing). Can still ungroup the item if its a single item selected and

PR #1571, opened 2025-05-10
Release Toggle Track Pitch Envelope v1.0

grayson-solis:

PR #1611, opened 2025-07-04
Release Toggle Track Pitch Envelope v1.0

grayson-solis:


Envelopes/Takes/graysonsolis_Toggle Track Pitch Envelope.lua
```diff
@@ -0,0 +1,97 @@
...
+
+local param = 3  -- Shift (cents) YOU CAN CHANGE THIS!!!!
+
+local env = reaper.GetFXEnvelope(track, fx_idx, param, true)
+if not env then
+  reaper.ShowMessageBox("Failed to access envelope.", "Error", 0)
+  reaper.PreventUIRefresh(0)
+  return
+end
+-+
```
cfillion: Stray `-+` here/syntax error.

Envelopes/Takes/graysonsolis_Toggle Track Pitch Envelope.lua
```diff
@@ -0,0 +1,97 @@
...
+if added then
+  show = true
+else
+  local v1,v2,v3 = chunk:match("VIS (%d+) (%d+) (%d+)")
+  show = not (v1 == "1" and v2 == "1" and v3 == "1")
+end
+
+chunk = chunk
+  :gsub("ACT %d+",         "ACT " .. (show and 1 or 0))
+  :gsub("VIS %d+ %d+ %d+", "VIS " .. (show and "1 1 1" or "0 0 0"))
```
cfillion: Feedback from the previous PR still apply: https://github.com/ReaTeam/ReaScripts/pull/1571/files#r2083305666

(This is not a packaging problem though so won't block a release here. Just a heads up for likely bugs.)

cfillion (changes requested):

Envelopes/Takes/graysonsolis_Toggle Track Pitch Envelope.lua
grayson-solis: Disabling that setting is the only way I got it to work. might be another way but it has worked fine for me for the last month or 2 ive been using it hours daily

Envelopes/graysonsolis_Toggle ReaPitch shift (cents) envelope.lua
```diff
@@ -0,0 +1,96 @@
...
+-- @about
+--   Toggles Shift (cents) envelope lane for ReaPitch on selected track. Adds ReaPitch if missing. Similar to the "volume" or "pan" default track envelopes.
+--   The parameter used can be toggled by editing the lines below!
+
+reaper.PreventUIRefresh(1)
+reaper.Undo_BeginBlock()
+
+local track = reaper.GetSelectedTrack(0, 0)
+if not track then
+  reaper.PreventUIRefresh(0)
```
cfillion, line 13: By the way, while this line here (and same a bit below) is ~~correctly undoing~~ the `PreventUIRefresh` from above, it forgets to the the same for `Undo_BeginBlock`.

An easy solution would be to wait until later before enabling `PreventUIRefresh` and `Undo_BeginBlock`. :wink: 

Thankfully REAPER has a fail-safe against scripts doing this kind of mistake so there are no ill effects.

Buy-One: Schwa suggests here https://forum.cockos.com/showpost.php?p=2297817&postcount=2
> You should always call PreventUIRefresh(1) to initiate and **PreventUIRefresh(-1)** to end

Is 0 a valid argument as well?

cfillion: My bad, nope it is not. (Well, not "invalid" per se, but adding 0 just does nothing.)

PR #1619, opened 2025-07-16
Release Kanban v1.0.2

JulyDay0788:
### Added
   - Deadline and completed time stamps in 'To Do' and 'Done' colums respectively
   - Different color for ovedue tasks in 'To Do' column
   - Arbitrary new line support in task text
   - Improved task box click animation
   - Tasks moved between columns are inserted into the new column at the mouse cursor position
   - Header right-click menu with options to chronologically sort tasks in 'To Do' and 'Done' 
      columns, restore default window size, batch delete tasks, jump to top/bottom/left/right edge,
     export/import tasks
   - Verticial scrollbar has been redesigned into a swipe track which doesn't have 
     a thumb and doesn't depend on window size. Scroll amount depends on the
     distance between initial and final mouse curror y coordinate
   - Horizontal scroll swipe track
   - Storage and recall of window size, coordinates, dock state and scroll position.
     When switching between project tabs only the scroll states are recalled.
   - Restoration of the window focus when task dialogue closes or tasks are imported
   - Prevention of window resize beyond maximum dimensions
   - Clock in the header

### DIFF
--- a/Project Properties/jd_Kanban.lua
+++ b/Project Properties/jd_Kanban.lua
@@ -1,6 +1,25 @@
 -- @description Kanban
 -- @author Julyday
--- @version 1.0.0
+-- @version 1.0.2
+-- @changelog
+--   ### Added
+--      - Deadline and completed time stamps in 'To Do' and 'Done' colums respectively
+--      - Different color for ovedue tasks in 'To Do' column
+--      - Arbitrary new line support in task text
+--      - Improved task box click animation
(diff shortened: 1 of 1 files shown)


Buy-One:
Well, i didn't mean this version to be public, i submitted it to your repo for your review as i indicated in the pull request comment, because it's littered with comments so you could identify new code for easy comparison with the old.

If i knew of a way to edit the submitted file, with your permission i could overwrite the current code with the clean version.

PR #1637, opened 2025-09-19
Release ReaEQ Will Putney Preset v1.0

Lanndyla:
First Release/Test


cfillion:
This would be better shared on https://stash.reaper.fm/ as a Track Template or FX preset file rather than as a script.

PR #1661, opened 2025-11-03
Release EmojImgui v0.3 (Initial release)

BenTalagan:
First version of EmojImGui library. I've committed some src files to build the fonts and their specifications even if they won't be checked out by reapack, as well as a little doc on the process, I thought it was better to store everything at the same place just in case this needs to be maintained.

Rendering a few thousands of icon buttons is not extremely light on the CPU (it still is fluid), but I don't really see how to do it better, and luckily the picker won't stay long on the screen anyway.


cfillion:
index.xml should not be modified.

As this is a release repo (not for dev), unpackaged files don't belong here. I can merge them for now anyway, but note that they'll get lost when moving out of GitHub for ReaPack v2.

BenTalagan:
Ouch this is accidental (for index.xml). It should not have been pushed. Regarding other files, I can move them to another "build" repo, no problem with this.

BenTalagan:
I can prepare another, cleaner PR with only packaged files, fix the index.xml issue and move build files to another repo, to prevent you from doing the tedious job.

PR #1664, opened 2025-11-09
Release Workflow scritps v1.0

hidearmor:
Version 1.0.
First package I have submitted.
Going to make a youtube-guide and upload link later.


cfillion (changes requested):
Each unrelated action should be released as a separate script package so they may be installed/updated individually.

Action filenames should use sentence case instead of PascalCase.

A few files were left with the default (eg. "New file 6.lua") storage name.

hidearmor:
OK, I'll drop this PR and make separate ones :)

PR #1670, opened 2025-11-12
Release Smart Rename v1.0

grayson-solis:


Buy-One:
> It CANNOT rename envelopes / envelope lanes. Sorry I wasn't able to figure this one out!

Built-in envelopes cannot be renamed. FX parameter envelopes can by renaming the actual parameters via chunk or via API and action, e.g.

    -- set parameter associated with envelope as last touched
    reaper.TrackFX_SetNamedConfigParm(track, fx_index, 'last_touched', param_index) -- param_index is string
    -- call the dialogue
    reaper.Main_OnCommand(41145, 0) -- FX: Set alias for last touched FX parameter

Lanes can be renamed via action `Track lanes: Rename lane under mouse`

grayson-solis:
This code already handles lanes. I already tried the solution you suggested with envelopes and unfortunately, it doesn't work! I tried a bunch of other things too. For whatever reason, it just doesn't seem possible. I'm cool to go ahead and push it without this functionality though since I feel like I tried everything already

Buy-One:
True, it only works for parameter part of the envelope name and to get rid of the FX instance part of the name the FX instance own name must be made blank which is impractical because it'll be removed in the FX chain as well.

PR #1672, opened 2025-11-13
Release Spectral Edits Preset System v1.00

amagalma:


cfillion:
- Script package filenames should use sentence case
- This should probably go in a new Spetral Edit category instead of Various?

amagalma:
> * Script package filenames should use sentence case

Sorry, I didn't know! Can you rename it?

>     * This should probably go in a new Spetral Edit category instead of Various?

I don't know.. It could.. Should I create such a directory? (I have no idea how)

PR #1695, opened 2025-12-12
Update Reannotate v0.4.4 > v0.4.5

BenTalagan:
Changelog

 - [Feature] Stickers can be positioned top or bottom
 - [Bug Fix] Markdown is now wrapped in posters
 - [Bug Fix] Plain text is now wrapped in posters


BenTalagan:
The main file had a wrong version number and I forgot to save the correction before submitting - hence the other PR.

PR #1701, opened 2026-01-04
Release Perfect Timing! - Audio Quantizer v0.40

80icio:
- Improved transient detection with more accurate peak and RMS measurements, 
  DC offset correction, and smarter retrigger behavior
- Major speed improvements for filters (lowcut, hicut, transient, gain) - 
  up to 2x faster processing
- Fixed incorrect grid drawing in the internal visualizer


80icio:
Version 0.4 was merged but the number is considered < 0.32, so I changed it to 0.40, that should work

PR #1706, opened 2026-02-09
Release MBCleaner v1.0

vovvee:


cfillion:
Public ReaImGui scripts must enable backward compatibility. `get_file_mod_time` is broken on Windows (lfs is not a thing in ReaScripts) and with GNU `stat` in $PATH (aka Linux). It's also vulnerable to command injection.

PR #1711, opened 2026-03-26
Release Quad Cortex MIDI control v2.3.2

6wheels:
- Fix reapack-index linked to wrong git commit
- README update with region combination name examples
- Add info after setup on how to add a toolbar button for easy access, or re-execute the setup wizard if needed.
- Add MIDI output ID validation in the setup wizard to prevent invalid entries.
- README update with clearer setup instructions and troubleshooting tips.
- Add info after first config on how to add a toolbar button for easy access, or re-execute the setup wizard if needed.
- Add setup to list action list (for easier access and reconfiguration)
- Fix folders to fix category detection with reapack-index
- Split from original Quad Cortex MIDI Control script to create a more modular and maintainable codebase.
- Added a setup wizard for first-time configuration.
- Improved logging and error handling.
- Automatically creates the MIDI output track if it doesn't exist.


cfillion:
The action file does not need to be moved to the new subdirectory (doing so erases any user shortcuts/toolbars/custom actions). Also the subfolder should have the same author prefix ("bertrandc_Quad Cortex MIDI control") to make obvious which package it belongs to.

The instructions mention the ability to re-run the setup wizard, however setup.lua is not registered in the Action List. If registering it, then it should also be named in sentence case with the author prefix (eg. "bertrandc_Quad Cortex MIDI control (setup).lua").

6wheels:
My idea for moving files to a subdirectory was to keep files organized in the repository.
Ok for the main script and the setup, but how can I named the lib.lua, also with the author prefix? bertrandc_lib.lua?

cfillion:
lib.lua is OK as-is since it's not user-visible (and is in the clearly named directory for repo management). The author prefix and sentence case requirements are for the Action List.

6wheels:
Hi Christian,
I updated the pull request, is it ok now?

PR #1712, opened 2026-03-30
Release ReaLoop Arranged Live Looper v2.0

ilovemedia:
In an earlier version, the script was called ‘Arranged Live Looping’.
- Complete engine rewrite: ultra-optimized single defer loop with O(1) marker-track caching.
- Introduced predictive lookahead for cross-track recording transitions.
- Real-time MIDI data throttling to prevent performance stutter, avoiding O(N) evaluations in large projects.
- Real-time deterministic timeout tracking for ghost-take audio cleanup.
- "Auto-kill" logic terminates script cleanly when playback halts.
- Automatic Loop Splitting: Slices dragged-out loop items cleanly into loops at script initialization.
- Miniscule (5ms) linear crossfades applied to playback audio automatically to prevent waveform clicks.
- Added 'monitor' naming keyword to dedicate tracks to input-pass-through rather than recording.
- Added 'manual' naming keyword to allow per-track suppression of auto-fades.
- Added alias support explicitly allowing markers to fire track switches for tracks prefixed with 'manual'.

### DIFF
--- a/Various/ilovemedia_ReaLoop Arranged Live Looper.lua
+++ b/Various/ilovemedia_ReaLoop Arranged Live Looper.lua
@@ -1,0 +1,797 @@
+-- @description ReaLoop Arranged Live Looper
+-- @author ilovemedia
+-- @version 2.0
+-- @changelog
+--   In an earlier version, the script was called ‘Arranged Live Looping’.
+--   - Complete engine rewrite: ultra-optimized single defer loop with O(1) marker-track caching.
+--   - Introduced predictive lookahead for cross-track recording transitions.
+--   - Real-time MIDI data throttling to prevent performance stutter, avoiding O(N) evaluations in large projects.
+--   - Real-time deterministic timeout tracking for ghost-take audio cleanup.
+--   - "Auto-kill" logic terminates script cleanly when playback halts.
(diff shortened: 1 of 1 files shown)


cfillion:
The current PR adds a new package instead of updating "Arranged Live Looping" from v1.0 to v2.0 (losing its history).

To load an existing package for editing: 1) type its name in the "Display Name" field 2) click on it in the list that pops up.

EDIT: (ReaPack will gain the ability to rename packages without deleting the previous one + creating a new one in the upcoming v2.)

ilovemedia:
Thanks. That renaming tool will be very useful. I made the change because users typically search for 'looper' rather than 'looping'. The old version is no longer functioning reliably and introduces some issues, so it could be marked as deprecated. Have a nice day!

ilovemedia:
Ok, I'll remove v2.0 and the changelog, and submit a new pull request with the new name (v1.0).

ilovemedia:
Do you know when the new release ReaLoop Arranged Live Looper v1.0 (#1714) is expected to be merged?
I'm currently preparing the video tutorials, and the only remaining section covers the installation process and how to add it to the toolbar. I would prefer to wait until it is available in the ReaPack repository before recording that part.
Thank you.

PR #1715, opened 2026-04-04
Release Turing Complete Web Interface v0.alpha2

rexendevar:


rexendevar:
The largest file here is actually an html webpage so keep that in mind :+1:

cfillion:
User-accessible actions (even if meant for use by the web interface in a headless setup, they're still visible in the Action List) must use the "authorprefix_Sentence case name.lua" naming convention.

Considering REAPER comes with "more_me.html", I think it would be best to follow the same naming scheme for web interfaces. How about "rxfx_turing_complete.html"?

"Web Interface" will be the category in ReaPack (searchable in v1) and in v2 category is selectable like below in the browser so repeating in the package name is redundant. How about simply "Turing-complete" as the package's display name?

EDIT: Or "Rek's turing-complete controller" to match the `<title>`?

20260406_14h19m16s_grim

I'll update the script command IDs accordingly and add the required ttf file in a subfolder. Will wait for you to double-check the changes before merging.

cfillion:
20260406_15h01m51s_grim

rexendevar:
Everything should be set up now.

PR #1719, opened 2026-04-21
Release Gamepad Controller for Reaper v1.0

rexendevar:
Initial upload


cfillion:
Is it called "Gamepad Controller for Reaper" or "Hotpluggable Gamepad Controller"? The package name should match the action name (and use sentence case not Title Case).

PR #1720, opened 2026-04-25
Workaround corner case involving "current position"

rexendevar:


Various/rxfx_Gamepad controller for REAPER.eel
```diff
@@ -1,7 +1,7 @@
 // @description Gamepad controller for REAPER
 // @author Rek's Effeks
-// @version 1.1
-// @changelog Added Windows support and trigger scrubbing
+// @version 1.1hotfix
```
cfillion: Letters in the version number makes it a pre-release before 1.1.

PR #1726, opened 2026-05-03
ReaPad: more fx controls + cut/dupe selection + (Turing-Complete) save button

rexendevar:
I really hope I'm running out of things to do like I think I am.


rexendevar:
I know it's really weird to change two separate packages in one PR but lemme explain. ReaPad & Turing-Complete are designed to be used together. ReaPad has a regular save button, but in cases where a project hasn't been saved yet, I need it to save in a very  specific file structure so Turing-Complete can also use it. So I made the autosave-as function part of Turing-Complete instead of ReaPad-only users having to deal with the same bizarre file structure

PR #1734, opened 2026-06-04
Release Reaper Web-Video Downloader v1.0

CalaFlan:
First Commit

### DIFF
--- a/Video/CalFlan_VIDEO_DOWNLOADER_Video.lua
+++ b/Video/CalFlan_VIDEO_DOWNLOADER_Video.lua
@@ -1,0 +1,129 @@
+-- @description Reaper Web-Video Downloader
+-- @author CalFlan
+-- @version 1.0
+-- @changelog First Commit
+-- @provides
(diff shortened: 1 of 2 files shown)


cfillion:
This uses os.execute/io.popen in a way that is both vulnerable to command injection and can easily freeze REAPER for >1 secs, plus requires externally-sourced files to function. As per the [guidelines](https://github.com/ReaTeam/ReaScripts/wiki/Guidelines) this cannot be released here in the current form.

(Unrelated to releasing, just an observation: by the time importVideo is invoked there is no guarantee a newer file hasn't been created after yt-dlp's, leading to get_last_file_name returning an incorrect value, or yt-dlp just didn't create any file leading to the same)

CalaFlan:
Yeah that’s all fair enough, I'll admit its a pretty scuff setup atm. I'm going to take a look and see if I rejig this using python in a way that isn't so dodgy. I'll submit something if I can get it working, either way thanks for taking the time to review this.

PR #1742, opened 2026-07-21
Update: 2026-07-21 20:35

flaneurette:
21-07-2026
==========

+ Fixed the listing of drum patterns
+ Fixed the freezing on Linux (only drew upon closure)
+ By default the chord section is now always checked, unless manually unchecked.
- removed one copyrighted drum pattern (was from a unknown? song)
- no longer supporting the windows binary .exe installer.


flaneurette:
Not sure why it fails... also got Github errors.

cfillion:
The errors are due to the removal of the required `@noindex` lines (to mark .py and .txt files as non-packages).

The main package file (MIDI Editor/Flaneurette_ReaChorder.py) needs to be updated (version & changelog) in addition to the code update, rather than moving it to MIDI Editor/Flaneurette_ReaChorder/ReaChorder.py.

The changelog should be in that file (just the new version's as ReaPack keeps history itself) rather than a changelog.txt. Screenshots should be listed in the `@screenshot` tag (same syntax as `@link` and `@donation` already present).

README should be moved to the package file's `@about` tag. VERSION.md doesn't apply here but if you want it to be included in the ReaPack release, it needs to be added to the list of files in the package file's `@provides` tag.

PR #1745, opened 2026-07-22
Release GuitarChorder v2.1 and ReaChorder v2.3

flaneurette:
- De-duplicated per-letter GUI/logic code
+ fixed freezing on Linux.
+ AI peer reviewed for bugs.


MIDI Editor/Flaneurette_GuitarChorder.py
```diff
@@ -1,10 +1,12 @@
 # @description GuitarChorder
 # @author Flaneurette
-# @version 2.0
-# @changelog update
+# @version 2.1 (refactored)
```
cfillion: This is equivalent to 2.1.refactored. Letters in a version number make it a pre-release (= existing users won't receive it unless they opt-in).

